PR #159, opened 2014-10-13
Fixes #138, disallow special characters for repository name

### LINKED ISSUE #138: Creating an repository with a . at the end makes it unaccessible
create a repository with a . at the end, and try to view/edit it..

ghost:
Fixes #138 
Furthermore repos with special characters cannot be deleted.

The class FileNameAttribute can be deleted (only usage). If Regex check the filename there can't be any characters (depends on Regex) which are disallowed in path.
FileNameAttribute should stay to handle errors if the regex will extended unwary.

The message for other Languages should be updated. Resource-Name is 'Validation_FileName_Regex'


alexkuznetsov:
After this fix we can't edit several ours repos. They have a dots in names - name like namespace in C# (Blah.Blah2.Blah3 or Blah.BlahBlah).
Wthat about review this pull request?

PR #193, opened 2015-01-07
Show tags on commits list

heringeidaniel:
This makes visible the tags on every commit


SoonDead:
I like this for the following reason: 

 * It is currently impossible to properly list commits between 2 tags.
 * Listing commits between 2 tags can funtion as an auto-generated changelog.
 * This pull request mitigates this. It is at least possible to see where the tags are on the "Commits" list.

PR #197, opened 2015-01-09
Regex fix

### LINKED ISSUE #194: Can't edit repository after "disallow special characters for repository name pull request"
Hello!
I repeat my comment from #159.

After #159  we can't edit several ours repos. They have a dots in names - name like namespace in C# (Blah.Blah2.Blah3 or Blah.BlahBlah).

What about to use something like that:
```csharp
var repositoryName = repositoryName.Trim('.');
```

It will be removed dots from start and end of repository name.

### LINKED ISSUE #198: Collection for "special" repo names
This is not really an issue, but i would like to make a collection thread (like a forum) for the different repo-names.

In #194 it's an error that the dots are not allowed.
In PR #197 i want to fix it that Repos can habe words, digits, dot, minus and underscore.
Repo names like the following would be not working:
abc. (because dot in the end)
!
!a
a!
"%§
And something else..

I think it could be helpful to collect some "special" names you use, to prohibit a problem like #194.
With the PR some supported names are:
Projectname
Project_ABC
_Project
1Project
Name.Space.Project
Name-Space.Project
abc-def_ghi.jk

If there are no problems with the names, the PR can be done. Otherwise i will update the regex, so we can use several more characters.

mbedded:
Fix #194 
Fix #198 

Update the regex. It uses groups (\w) instead of custom groups (a-z, A-Z, 0-9).
Added the "dot" in the tetx for the error-messages.

Regex checks if last sign is a word (special characters like !"§$%*' are not allowed).


mbedded:
We should wait for some responses in #198.
I want to prevent a Problem like #194.

PR #199, opened 2015-01-12
Change commits layout.

whosa:
--------------
raw like this
--------------
raw

--------------
current(don't show msg in link tips)
--------------
current


jakubgarfield:
That's a really nice piece of functionality that is missing in the current implementation. Wouldn't it be better to have the full commit message in the commit detail view instead of commits list?

I'd prefer having the logic responsible for parsing the commit message outside the view. If you'd go with the display in the details, it would be in RepositoryBrowser class and the method GetCommitDetail - ToModel would be a perfect place for this decorating method. You need to extend RepositoryCommitModel with a new field as well. After that, you can just add this field to the view.

Do you think it's reasonable?

whosa:
Hi jakubgarfield:
    I change the implementation like you said. Now, it like this
list
image

PR #202, opened 2015-01-15
Wrapped group identity translation in try/catch and filtered out null tr...

jshepler:
...anslations

Some group identities are not translatable and throws an exception, "System.Security.Principal.IdentityNotMappedException: Some or all identity references could not be translated."

I don't think there is a way to detect if an identity is translatable before attempting it, but we have this issue in our domain. I've run into this error before and I think the only thing that can be done is wrap the translation in a try/catch and eat the exception.

### DIFF
--- a/Bonobo.Git.Server/WindowsIdentityImporter.cs
+++ b/Bonobo.Git.Server/WindowsIdentityImporter.cs
@@ -64,10 +64,13 @@ namespace Bonobo.Git.Server
                     IdentityReference translated = null;
                     try
                     {
-                        translated = @group.Translate(typeof(NTAccount));
+                        if (@group != null)
+                        {
+                            translated = @group.Translate(typeof(NTAccount));
+                        }
                     }
 
-                    catch { }
+                    catch (System.Security.Principal.IdentityNotMappedException) { }
 
                     return translated == null ? null : translated.ToString();
                 })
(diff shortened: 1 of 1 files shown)


jakubgarfield:
Is it possible to catch just NotMappedException instead of global catch block?

jshepler:
Sure it is and normally I would have allowed other exceptions through or tried to fit into any existing exception handling. But in this use case, failing to translate a group isn't important enough to prevent the user from being imported, crashing the request, and throwing up a ysod - making the app completely unusable to that user. Therefore, I believe it's appropriate to eat any exception from this particular call to Translate. Unless you want to log the exception, then a single catch of System.Exception is sufficient.

jakubgarfield:
From my general point of view it's ok to try to catch a specific exception and do nothing around this specific call. But it's not ok to catch everything - even StackOverflowException, OutOfMemoryException, and others.

What I'm suggesting is doing something like

===
catch(System.Security.Principal.IdentityNotMappedException) { }
===

instead of the 

===
catch {}
===

jshepler:
I completely understand what you were suggesting. This is my first pull request, so I don't know if there's an etiquette I'm not following. Were you asking me to make that change or are we having a discussion?

jakubgarfield:
No worries :) if you agree that it make sense to change it and I'll merge it.

jshepler:
Not really, but not enough to argue about. Go for it.

PR #216, opened 2015-02-27
Feature: Commit messages like in github

alexkuznetsov:
Hello!

I refactor commit messages, now they look like in github.
I have the trouble with storing splitted commit messages, i use for it properties Message and MessageShort. 
Any ideas for refactor, suggestions?

Screenshots:
Default state:
screenshot_1
Expanded state:
screenshot_2

### DIFF
--- a/Bonobo.Git.Server/Bonobo.Git.Server.csproj
+++ b/Bonobo.Git.Server/Bonobo.Git.Server.csproj
@@ -204,6 +204,8 @@
     <Compile Include="Git\GitService\ReceivePackHook\ReceivePackCommitSignature.cs" />
     <Compile Include="Git\GitService\ReplicatingStream.cs" />
     <Compile Include="Helpers\MembershipHelper.cs" />
+    <Compile Include="Helpers\RepositoryCommitModelHelpers.cs" />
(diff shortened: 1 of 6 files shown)


jakubgarfield:
Thanks for the PR, I need to review it more closely and hopefully will manage to do it by the end of the week.

alexkuznetsov:
@jakubgarfield thank you for response. As i say, there are some tricks in my solution. 
I am open for any recommendation for refactor.

jakubgarfield:
Nice work, personally I'd be happy with just using string.Substring instead of breaking based on the words size.

And having a decorator instead of helper + model would be also my preferred choice. Something like

```
CommitMessageDecorator
   ctor(message);
   prop String Title;
   prop String Details;
```

but your implementation is working and the feature is really nice :)

alexkuznetsov:
@jakubgarfield, thank you for the merging!

I will try to implement the approach with decorator. 
The solution with breaking based on words size for me is more good instead using of subscring - all words are **comleate** in that case.

whosa:
I found XSS in this pull request. If you write javascript in commit message , it will be run.

Issue #221, opened 2015-03-10
XSS in pull request #216


whosa:
I found XSS in pull request #216 . If you write javascript in commit message , it will be run.

alexkuznetsov:
hmm. I will check it.

alexkuznetsov:
@whosa what do you think about #222?
Can you test it for xss?

PR #222, opened 2015-03-10
XSS fix

alexkuznetsov:
Hello!
I check my code and rewrite the helper class. So #221  now fixed, i think.

Tests for xss took [there](https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet).

Can you review my commit?


alexkuznetsov:
Screenshot with results:
screenshot_1

PR #227, opened 2015-03-18
added configuration loading for mail send

EmaDev:
Added configuration loading from web.config on
<configuration> => <appSettings> for
smtpHost
smtpPort
smtpUsername
smtpPassword
smtpSenderAddress


latop2604:
Why not use the standard system ? https://msdn.microsoft.com/en-us/library/ms164240(v=vs.110).aspx

alexkuznetsov:
As for me, @latop2604 is right. Standard system is more clear and working good.

EmaDev:
Oh I'm sorry, it's correct, 
I don't knew this method! 
Thanks

Issue #77, opened 2014-01-28
Browsing / acessing  public/anonymous repositories w/o credentials


moljac:
Hi

It seems that showing listing anonymous/public repos without authentication / authorisation is not possible. Or am  I wrong?

What would be the bet way to do it? On Login screen to add list of public repos?

thanks

mel

hyuan:
Are you still planning to implement this feature? There is a checkbox "Anonymous" on the repository setup page, but seems it's not functioning.

RedX2501:
@jakubgarfield I think this has been done already? Can we close this?

moljac:
If it is implemented. Sure. I had no time to look at it.

Sorry.

jakubgarfield:
This one is not implemented, you have anonymous access with git, but not from UI interface.

jakubgarfield:
@glacasa Did you manage to get any progress with this feature?

glacasa:
No sorry, I didn't work on this project for a long time

PR #232, opened 2015-03-19
Anonymous listing - issue #77

glacasa:
I updated the index page to allow anonymous user to browse directories.
Here are my modifications :

Added a checkbox in the global settings to allow anonymous browsing (I'm not sure if this should be an option or not, because we only display repositories already flagged as public)

Changed the Authorize on `Home/Index` and `Repository/Index` : if the setting defined above is checked, the page is allowed.  
At this point, only the public repositories are listed, I didn't have to do anything to hide the private ones.

In order to browse the repository content, I had to update the `WebAuthorizeRepositoryAttribute` : I check if the repository is public (AllowsAnonymous) before to check if the user is logged.


jakubgarfield:
I don't think the option in the web config is necessary.

Also, this change doesn't make sense if you don't allow to browse public repositories (details, view commits, etc) and I think it can be done only by modifying the WebAuthorizeAttribute. There is no need to introduce a new one - actually it might complicate things a little.

PR #235, opened 2015-03-22
added links to the committed changes so we can reference them via links.

spoiledtechie:
added links to each committed change so I can reference them.

### DIFF
--- a/Bonobo.Git.Server/Content/Site.css
+++ b/Bonobo.Git.Server/Content/Site.css
@@ -192,6 +192,3 @@ nav.branches ul li:hover ul li { float: none; }
     top: -9px;
     vertical-align: middle;
 }
-
-.pull-right { float:right; }
-.pull-left {  float:left;}
--- a/Bonobo.Git.Server/Content/Site.css
+++ b/Bonobo.Git.Server/Content/Site.css
@@ -192,3 +192,6 @@ nav.branches ul li:hover ul li { float: none; }
     top: -9px;
     vertical-align: middle;
 }
+
+.pull-right { float:right; }
+.pull-left {  float:left;}
--- a/Bonobo.Git.Server/Models/RepositoryModels.cs
+++ b/Bonobo.Git.Server/Models/RepositoryModels.cs
@@ -103,6 +103,7 @@ namespace Bonobo.Git.Server.Models
 
     public class RepositoryCommitChangeModel
     {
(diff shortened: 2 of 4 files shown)


Bonobo.Git.Server/Content/Site.css
```diff
@@ -192,3 +192,6 @@ nav.branches ul li:hover ul li { float: none; }
     top: -9px;
     vertical-align: middle;
 }
+
+.pull-right { float:right; }
+.pull-left {  float:left;}
```
jakubgarfield: Remove unused CSS

Bonobo.Git.Server/Views/Repository/Commit.cshtml
```diff
@@ -3,11 +3,13 @@
 @{
     ViewBag.Title = Resources.Repository_Commit_Detail;
     Layout = "~/Views/Repository/_RepositoryLayout.cshtml";
+    var commitDetails = Model.Changes.Where(i => i.Status == LibGit2Sharp.ChangeKind.Modified || i.Status == LibGit2Sharp.ChangeKind.Added).ToList();
 }
-@helper HeaderForDiff(RepositoryCommitChangeModel item)
+@helper HeaderForDiff(RepositoryCommitChangeModel item, int changeId)
```
jakubgarfield: Instead of modifying the view, add Id to the RepositoryCommitChangeModel and you can keep the view almost as it was.

Bonobo.Git.Server/Models/RepositoryModels.cs
```diff
@@ -177,7 +177,7 @@ public string Message
         public string TagName { get; set; }
 
         [Display(ResourceType = typeof(Resources), Name = "Repository_Commit_Changes")]
-        public IEnumerable<RepositoryCommitChangeModel> Changes { get; set; }
+        public List<RepositoryCommitChangeModel> Changes { get; set; }
```
jakubgarfield: No need to change the contract if you add Id property to the model.

jakubgarfield:
Happy to merge, if you move Id to the model and keep the view simpler. The generation of Change model is at https://github.com/jakubgarfield/Bonobo-Git-Server/blob/master/Bonobo.Git.Server/RepositoryBrowser.cs#L317

spoiledtechie:
updated code to work as you said.

PR #243, opened 2015-04-03
fix db update for sql server

alexkuznetsov:


alexkuznetsov:
I think [nvarchar](http://stackoverflow.com/questions/144283/what-is-the-difference-between-varchar-and-nvarchar)would be better choice for name

jakubgarfield:
There is no difference between varchar and nvarchar in sqlite, is this a fix for a different database? Also, how about users that already migrated?

alexkuznetsov:
All my changes are related only to a SQL Server.
Automatic migration to last version with SQL server as database impossible, because alter script is incorrect.

PR #247, opened 2015-04-13
update font awesome, add file info display

lkho:
Sorry for the translation update again, the translation of "Normal View" in Chinese is the best match of "Blob" in that situation.
May be the file info can include sloc also...


jakubgarfield:
Front end dependencies (actually just font awesome) are managed with bower. Could you use it for font awesome update as well? You should just change bower.json inside project root and run bower install.

jakubgarfield:
Happy to merge the other two commits though.

Issue #254, opened 2015-04-26
Error in highlight.js = Preview of this file is not supported.


padremortius:
Bonobo showed message "Preview of this file is not supported." when we have error in highlight.js (not correct detect template). For correction need make changes in FileDisplayHandler.GetBrushes and rebuild application. Maybe should make a correspondence table extension to the template to an external configuration file or database? This will allow to make changes without rebuilding project.

padremortius:
For example, get file https://github.com/project-jedi/jvcl/blob/master/jvcl/run/JvAnimate.pas 
Change extension from *.pas to *.pp. Upload into git. Open this file in web-interface Bonobo. We see error 'Preview of this file is not supported.'. And push the button Raw (Error 404- File not found).

jakubgarfield:
Confirmed bug that needs fixing, it might be due to different encoding of the file. Any takers?

padremortius:
Return extention to *pas and we have not error.
I trying.

PR #257, opened 2015-05-02
Fix bug #254 with comment "Preview of this file is not supported". ext2a...

padremortius:
...pp.txt - correspondence table extensions and templates of highlights. Add new rule in .gitignore (add ext2app.txt to white list)


jakubgarfield:
Thanks for the PR. I thought about it, but I'd prefer hard coding the value in the existing C# file, instead of dynamically loading ext2app.txt.

The reasons are mostly security (opening a file on the server), performance (opening a file every time someone browse a file in repository browser) and I don't think the list is going to change a lot in the future - if there is something missing, let's just add it to the code and do the release.

Would you mind changing it?

padremortius:
Hardcode is not good idea. In real life we have identical extentions for different application (example, *.p, *.m, *.pl, *.rb and etc.)

1. I can transfer the contents of a file into the database, but then need to make an interface for editing associations. It's mostly security, but slowly.
2. I can place a bug on highlight.js, but it is not clear how to describe it with the possibility of repetition

PR #259, opened 2015-05-05
Fix issue #254

padremortius:


jakubgarfield:
Does the raw view still fail? Do you have an idea why?

padremortius:
> Does the raw view still fail?
No.
>Do you have an idea why?
No.

jakubgarfield:
It doesn't? wow, I don't understand :)

padremortius:
Very strange work highlign.js. In manual written that class="nohighlight" - turn off highlight... But not... Hightlight is still work. May be update this component?

PR #293, opened 2015-07-14
Remove origin branches after cloning

### LINKED ISSUE #288: When using clone you can't use branches cloned over.
Hi, 

I have seen a slight issue when cloning repositories.

Repo A has master and develop braches and some tags. 
When cloning Repo A and calling the new one Repo B master and develop branches are prefixed with origin/master and origin/develop. 

The problem now is that you cannot do anything with them and the actual fact is that you can only use the master branch that is recreated. 

This is Repo A 
repo a
This is Repo B 
repo b

Does anyone know a workout either to delete the origin/master and origin/develop branches from view or to make them available to use to be able to check them out? 

Thanks in Advance.

latop2604:
Fix #288

### DIFF
--- a/Bonobo.Git.Server/Controllers/RepositoryController.cs
+++ b/Bonobo.Git.Server/Controllers/RepositoryController.cs
@@ -454,6 +454,15 @@ namespace Bonobo.Git.Server.Controllers
                             };
 
                         LibGit2Sharp.Repository.Clone(sourceRepositoryPath, targetRepositoryPath, options);
+
+                        using (var repo = new LibGit2Sharp.Repository(targetRepositoryPath))
+                        {
+                            if (repo.Network.Remotes.Any(r => r.Name == "origin"))
+                            {
+                                repo.Network.Remotes.Remove("origin");
+                            }
+                        }
+
                         TempData["CloneSuccess"] = true;
                         return RedirectToAction("Index");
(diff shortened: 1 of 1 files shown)


jakubgarfield:
Can you please explain where the origin prefixed branches got there at the first place? I've never encountered this issue, so I don't quite understand.

latop2604:
Found in git-clone description : https://www.kernel.org/pub/software/scm/git/docs/git-clone.html
> This default configuration is achieved by creating references to the remote branch heads under refs/remotes/origin and by initializing remote.origin.url and remote.origin.fetch configuration variables.

In our case, the origin remote points to a local url like C:\\...

latop2604:
This pull request is just one of the two options : remove origin remote or hide remote branches in view.

jakubgarfield:
Oh I see, so the repo is originally cloned from a different location and Bonobo Git Server is not the place where it was created.

jakubgarfield:
Is there any chance that there might be an infrastructure setup with a hook that would update this first location as well? That some users might be dependant on those branches and by removing them we are limiting them?

latop2604:
No exactly. The removed remote origin will only be created during bonobo clone action. The origin remote, in this case, will be the local path of the source repository. So we can delete it safely.

Issue #189, opened 2014-12-21
Update git/msysgit for recent CVE-2014-9390 vulnerability


thekaveman:
See the [GitHub announcement](https://github.com/blog/1938-vulnerability-announced-update-your-git-clients) and the original [vulnerability announcement](http://article.gmane.org/gmane.linux.kernel/1853266) for more info. 

>The vulnerability concerns Git and Git-compatible clients that access Git repositories in a case-insensitive or case-normalizing filesystem. An attacker can craft a malicious Git tree that will cause Git to overwrite its own .git/config file when cloning or checking out a repository, leading to arbitrary command execution in the client machine. Git clients running on...any version of Microsoft Windows (NTFS, FAT) are exploitable through this vulnerability.

jakubgarfield:
#223

PR #297, opened 2015-07-19
Updated to msysGit-1.9.5-preview20150319

larshg:
I did the following to test it:

Created a new repository
Cloned it
Added a new file and commited and pushed it.

All went well - you mentioned some errors in #189 - how can I provoke these?


jakubgarfield:
I think that last time I did this simple test, it failed.

Try to run the regression test as described https://github.com/jakubgarfield/Bonobo-Git-Server/blob/master/Bonobo.Git.Server.Test/MsysgitIntegrationTests.cs

larshg:
I just tried to debug it a little and this are the changes i noticed:

CloneEmptyRepository:
Originally output is: Cloning into Integration...   - new output is "" ie empty string.
Originally error: warning: You appear to have cloned an empty repository. - new output is: Cloning into 'Integration'...
warning: You appear to have cloned an empty repository.

So it seems that they have moved all the output to the error output.

The output is also changed for the following subtests:
CloneRepository
PullRepository
Pulltag

How would you go about it? Want to make a if/switch depending on the git version and correct the "expected" output? or take it to msysgit git page and ask for clarification?

It seems that git 1.9.5-preview20150319 does what it supposed to, it "just" print out differently, ie. more/less data in either output or error stream.

jakubgarfield:
If/switch would be fine for this type of testing - there will be only one version of git on the server though.

larshg:
@jakubgarfield I have updated the test so 1.9.5 passes.
I changed it to use relative paths from debug/release folder.
I have only tested 1.8.1.2, 1.8.3 and 1.9.5.

jakubgarfield:
Awesome, nice work! :)

latop2604:
In the csproj, some git dll are referenced but not exits in the repo.
I do not found any error after removing broken reference. But I can miss something.

larshg:
@latop2604 I just recloned the repository and I didn't get any missing references - maybe the nuget restore didn't complete properly?

latop2604:
image

PR #303, opened 2015-07-30
Merge upstream

Duikmeester:


Duikmeester:
Somehow this always goes the wrong way :8ball:

PR #369, opened 2015-11-19
Added and translated missing Chinese resource keys

StarryLibra:
Appended chinese translation of missing resource keys.


StarryLibra:
Oh, click the worng button.

Issue #378, opened 2015-12-09
AD integration issues


mattepi:
I've encountered a few issues in using the Active Directory integration:

1. Nested group permissions don't work
2. Logging in without specifying a domain does not work throughout the system. The login occurs but the roles are not properly populated.
3. Username is case sensitive when loading the roles, which may be an issue for AD users who do not know the casing of their username in AD.

RedX2501:
@mattepi Since the merge of your PR is this still happening?

mattepi:
After applying my changes I no longer see the issues mentioned.

PR #379, opened 2015-12-09
Active directory updates:

mattepi:
1. Use nested groups for permissions
2. Allow logging in without specifying the AD domain (should use the default for all AD look-ups if one is not specified)
3. Update so username at login is not case sensitive when retrieving roles.


mattepi:
Here's the issue that was opened, https://github.com/jakubgarfield/Bonobo-Git-Server/issues/378.

PR #385, opened 2015-12-27
Increase button column width to prevent line break caused by long button texts.

mischalandwehr:
line-break

### DIFF
--- a/Bonobo.Git.Server/Views/Repository/Index.cshtml
+++ b/Bonobo.Git.Server/Views/Repository/Index.cshtml
@@ -27,12 +27,12 @@
 }
 
 <div class="pure-g">
-    <div class="pure-u-3-5">
+    <div class="pure-u-2-5">
         <h1>
             @Resources.Repository_Index_Title
         </h1>
     </div>
-    <div class="pure-u-2-5 add">
+    <div class="pure-u-3-5 add">
         @{
             var sortGroupQuery = Request.QueryString["sortGroup"];
 


jakubgarfield:
How about reducing the length of the Neues Repository erstellen? That would prevent breaking to two lines with the benefit of keeping valuable space for repository titles.

mischalandwehr:
IMHO the German resources are improvable. "Neues Repository erstellen" could shorten to "Repository erstellen" (you always create new things), "Repository Verwaltung" to "Repositories", ... Some translations are missing or faulty. So putting some love to the German resources could solve this breaking, too.

But a fix via shorting the title is only a fix for the current desktop version. If we want to extend it for tablets and smartphones we will get the same problem again. The "royal way" would be to increase the column width and than updating the German resources.

jakubgarfield:
On tablets and smartphones it should (in theory) just stack under each other as it is using the responsive grid, right? Maybe each button should have it's own column (1-5).

mischalandwehr:
If you place each button in 1-5 columns, the gap between them will be very large. The current small gap  nicer. 

To enable tablets you could change the classes to "pure-u-1 pure-u-md-x-5". This would break the buttons under the headline and all would look great.

jakubgarfield:
Which solution would you prefer? I'm happy to merge your current PR or wait for the responsive one.

mischalandwehr:
I would prefere to merge this PR.

Next week I will fix the German translations and after that the responsive one.

Issue #361, opened 2015-11-09
Switch repository icons from ugly jpg to nice png


carlosefonseca:
Uploaded repository icons are being converted to jpeg and they look awful, full of compression artifacts, and transparency is being discarded, with black on its place.

Please consider switching to PNG, where there will be no compression artefacts and transparency is  kept.

I'm using Bonobo Git Server (4.0.0.0)

RedX2501:
@mischalandwehr any chance of turning the the referenced commit into a PR?

mischalandwehr:
It's already done and merged to the master branch.

PR #386, opened 2015-12-27
Improve repository logo quality #361

mischalandwehr:
Change icon type from JPEG to PNG and add high density support for large uploads (width of 72 or wider)


jakubgarfield:
What happens with repository images already stored as jpegs?

mischalandwehr:
Good point. I will check this.

mischalandwehr:
"Old" JPEG logos will be untouched, but embedded as base 64 PNG. Under windows everything will look fine, but I know other systems have problems with wrong mime types.

To solve this I could extend the encoding by an image type detection (by magic numbers). So a JPEG would have the "image/jpeg" mime type and a PNG a "image/png" one.

Issue #394, opened 2016-01-11
Clarification on windows auth with internal membership DB


RedX2501:
The documentation on installation (https://bonobogitserver.com/windows-authentication/) says that if I use Internal Membership the data could become out of date. this means that upon first login the password is retrieved from AD?

What I'd like to provide to my users is a SSO (single sign on) experience, they get to use their Windows domain password everywhere but the repo access should be managed by Bonobo after the user has been authenticated.

This used to work flawlessly on 3.4.3 but with the new options I'm not sure how to configure this scenario.

Currently on a fresh install with authprovider=windows and membershipprovider=internal I'm not even able to log in. Trying with the default admin:admin leads me to a page error.

There used to be a key `<add key="ShouldImportWindowsUserAsAdministrator" value="false" />` which I could set to tell it to import the first user as an admin. Does this still work?

RedX2501:
To answer my own question since I'm providing PR #400 to fix this.

The support for windows auth with an internal db was broken.

Now you can set `ImportWindowsAuthUserAsAdmin` to get the same effect.

Issue #374, opened 2015-12-07
Change User Display When Creating/Modifying Repository


kars85:
Hi,

I'm wondering if someone might know where to look and modify what is displayed on the Create/Modify repositories page for user delegation as contributor or administrator. I've looked in the database tables, but their must be some ASP work being down to retrieve user values. I hope to set the display to be Lastname, FirstName instead of username

Our facility uses two factor authentication, with the Active Directory usernames setup as their smartcard UPN number, not their legacy firstinitiallastname. Administrative account still use a readable username starting with ap., but normal users are set to UPN.

Thanks!

![IMAGE](https://cloud.githubusercontent.com/assets/6616548/11629308/12a6e1ba-9cbd-11e5-8231-0ebc5a1fc957.png)

RedX2501:
Same as #295

Issue #295, opened 2015-07-15
Windows Authentication could retrieve User's name for better overview


RedX2501:
When executing any action which lists usernames in Windows Authentication mode with an Active Direct it only lists the username although more information is available. This makes editing them quite difficult as most companies usernames are non descriptive as can be seen in the screenshot.

![2015-07-15_09-50-20](https://cloud.githubusercontent.com/assets/929932/8693448/718c3ed0-2ad7-11e5-9188-69a0a87b221f.png)

Would it be possible to retrieve the Name or eMail Address and display it instead or additionally?

jakubgarfield:
So first of all, you'd have to introduce a concept of a display name - you need to have unique username, and I believe full name or emails don't have to be unique.

After that, you need to retrieve this information in WindowsImporter class https://github.com/jakubgarfield/Bonobo-Git-Server/blob/master/Bonobo.Git.Server/WindowsIdentityImporter.cs.

Also, there is no updating of this information, so unless some updating mechanism is added those information will get stale over time.

To get a friendlier name - have a look at https://msdn.microsoft.com/en-us/library/system.security.principal.windowsidentity(v=vs.110).aspx if you can get anything relevant from there.

PR #400, opened 2016-01-12
Fix #396 #295 #394 #374 Fix WindowsAuth with Internal DB

### LINKED ISSUE #396: Cannot use Authentication provider Windows with MembershipService Internal
With a fresh install (empty db) the users retrieved from the authentication provider do not seem to be inserted into the db. So when the Membership provider tries to fetch them it fails.

You also cannot log in as the default admin account as it is not a valid account in the domain.

The part that seems to be failing is in 

https://github.com/jakubgarfield/Bonobo-Git-Server/blob/8faa960af319bc89a5dfc8e93bbe1f081bec3c74/Bonobo.Git.Server/Owin/WindowsAuthenticationHandler.cs#L41

Maybe upon successfully authenticating the user it should be inserted into the DB if not yet present. This would also fix #394 allowing a simple SSO functionality. Maybe password change should be disabled when using Window as an authentication provider.

Pinging @Ollienator as I think it was him who introduced this code.

RedX2501:
Added option ImportWindowsAuthUsersAsAdmin to be able to import all users automatically as admins if wanted. The first user is always imported as admin.

Since we cannot store the Password (we never see it) added WindowsAuth also to Gitauthorization.

Usernames are now saved again with the domain separator in the name as that is how i see it on the back side.

The edit user page does not work at the moment. I'm lookin into it.

Hopefully this fixes #396 #295 #394 #374 without introducing too many new problems.

Pinging @Ollienator since he introduced most of the new OWIN auth part.

### DIFF
--- a/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
+++ b/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
@@ -66,7 +66,7 @@ namespace Bonobo.Git.Server
                         {
                             if (pc.ValidateCredentials(username, password))
                             {
-                                httpContext.User = new ClaimsPrincipal(new ClaimsIdentity(AuthenticationProvider.GetClaimsForUser(username)));
+                                httpContext.User = new ClaimsPrincipal(new ClaimsIdentity(AuthenticationProvider.GetClaimsForUser(username.Replace("\\", "!"))));
                                 allowed = true;
                             }
                         }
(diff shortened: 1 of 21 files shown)


RedX2501:
There are two remaining issues:

The personal repository links can be mangled if there is a \ in the username. I have no idea on how to fix this as the doman is required to log in with git.

When viewing a repository details it wll show the userid instead of a pretty name. This is because the RepositoryModel only gets UserIds. Maybe on a second PR this can be fixed.

PR #405, opened 2016-01-13
Fix #401 Cleanup all outdated scripts and configs.

### LINKED ISSUE #401: Package.cmd and deploy-apps.cmd is outdated
@jmagly introduced package.cmd and deploy-apps.cmd in 2014. With the new 5.0 version which no longer requires two interfaces, these need to be updated.

It would be interesting to know what process @jakubgarfield uses so we could automate that to use with the new AppVeyor integration.

RedX2501:


Bonobo.Git.Server/Web.Release.config
```diff
@@ -1,6 +0,0 @@
-﻿<?xml version="1.0"?>
-<configuration xmlns:xdt="http://schemas.microsoft.com/XML-Document-Transform">
-  <system.web>
-    <compilation debug="false" xdt:Transform="SetAttributes(debug)" />
```
latop2604: This is required

latop2604:
Web.Debug.config was deleted but not in Bonobo.Git.Server.csproj line 576

Issue #407, opened 2016-01-13
ADMembershipService throwing error when using build# 2f9a001


chrismdev:
I had to change this line:

using (UserPrincipal user = UserPrincipal.FindByIdentity(principalContext, username))

to this:

using (UserPrincipal user = UserPrincipal.FindByIdentity(principalContext, username.Replace("!",@"\")))

RedX2501:
How did you provide your Username for git? The intended way is domain\username

chrismdev:
I am using the ActiveDirectory membership service so I I only need to supply the ActiveDirectoryMemberGroupName to allow a list of users. Here is my config:

    <add key="AuthenticationProvider" value="Windows" />
    
    <add key="MembershipService" value="ActiveDirectory" />
    <add key="ActiveDirectoryDefaultDomain" value="xxxxxx" />
    <add key="ActiveDirectoryBackendPath" value="~\App_Data\ADBackend" />
    <add key="ActiveDirectoryMemberGroupName" value="GroupName" />
    <add key="ActiveDirectoryTeamMapping" value="Developers=GroupName" />
    <add key="ActiveDirectoryRoleMapping" value="Administrator=Domain Admins" />

RedX2501:
Oh I see. Can you please provide a PR?

chrismdev:
I just did a quick check and AD does allow you to have an ! in the username, so I think my fix is not correct.

I think the solution would be to only convert the first ! in the username. Yes will submit a PR.

RedX2501:
Although many services will allow ! in the usernames the PR I submited will not like them as I do a Replace("!", "\\") in many places. Maybe you can fix those too, to only replace the first occurrence?

chrismdev:
This is my first PR so please be gentle.

I changed all the Replaces() I could find but left the views alone.

RedX2501:
You should fix the replace in the Views to for consistency IMHO.

chrismdev:
Will do this as soon as I get a chance

RedX2501:
Just a heads up that this will probably be fixed by #426

RedX2501:
@chrismdev Can please check that this is fixed on latest master? https://ci.appveyor.com/project/jakubgarfield/bonobo-git-server/build/artifacts

# Please backup your Database as latest master will upgrade it to a format not compatible with 5.1.1.

chrismdev:
[...]

PR #412, opened 2016-01-14
Fix for issue #407

chrismdev:
Replace first occurrence of character

### DIFF
--- a/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
+++ b/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
@@ -11,6 +11,7 @@ using System.DirectoryServices.AccountManagement;
 using Bonobo.Git.Server.Security;
 
 using Microsoft.Practices.Unity;
+using Bonobo.Git.Server.Helpers;
 
 namespace Bonobo.Git.Server
 {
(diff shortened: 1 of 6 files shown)


kfarnung:
I would recommend changing "UserNameHelper" to be "StringExtensions" following the pattern here:

https://msdn.microsoft.com/en-us/library/bb383977.aspx

That will allow the code to look like:

username.ReplaceFirstOccurrence("!", @"\");

kfarnung:
Another question, what's special about the first occurrence?  I'm not familiar with the domain of the problem, but is there anything else distinct about the "first occurrence" that could make the replacement even more scoped?

chrismdev:
Hi, this is my first PR so I appreciate the feedback.

I don't have a reason for choosing a static instead of an extension method. You're right, it would be neater but I wasn't sure I should be polluting the string class with my own newbie methods.

The problem is that when using windows authentication the "\" causes problems in the UI and has been replaced with an "!".

My code was just to replace the "!" with a "\ "so that windows auth would work again.

kfarnung:
You should probably make the class static as well, it has no instance members.

The reason I was suggesting making it part of StringExtensions was that the method, as coded, has no username functionality.  If you internalized the characters to replace, then I think "UserNameHelpers" makes more sense, but in the current form it just seems like "StringHelpers."

The extension methods are nice because they only pollute the class when you have the namespace enabled and they are essentially just static methods, but the IDE/compiler helps them be contextual rather than hidden off in a class somewhere.

chrismdev:
Good point, will rewrite as an extension method.

PR #420, opened 2016-01-21
Fix #191 Updating highlight.js to 9.1.0. Adding various languages.

### LINKED ISSUE #191: Update Highlight.js to latest version for better C/C++ language support.
The current version of Highlight.js used has poor support for C and C++. Updating to Highlight.js 8.4 greatly improves the support for C/C++.

RedX2501:
I added all the available languages from the website.


jakubgarfield:
When I cloned the latest version and compiled I have 

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        Bonobo.Git.Server/Content/components/highlight/.bower.json
        Bonobo.Git.Server/Content/components/highlight/.gitignore
        Bonobo.Git.Server/Content/components/highlight/AUTHORS.en.txt
        Bonobo.Git.Server/Content/components/highlight/AUTHORS.ru.txt
        Bonobo.Git.Server/Content/components/highlight/docs/
        Bonobo.Git.Server/Content/components/highlight/src/
        Bonobo.Git.Server/Content/components/highlight/tools/
```

Is there something missing in the .gitignore or should those files be added to the project?

RedX2501:
These are leftovers from you building highlight the first time. Now we have the precompiled version from the website.

larshg:
@RedX2501 How did you add the new highlight? bower?

I recently pulled master and I get an error with a reference to the github.css in highlight/styles - it seems that when I launch VS I restores the bower packages, which states that we are using highlight 8.0.0 [here](https://github.com/jakubgarfield/Bonobo-Git-Server/blob/master/Bonobo.Git.Server/bower.json#L6) and hence overwrite this commit.

Also there is a path [here](https://github.com/jakubgarfield/Bonobo-Git-Server/blob/master/Bonobo.Git.Server/App_Start/BundleConfig.cs#L20) which needs to corrected too - if this is the right way to update to highlight 9

What do you think @jakubgarfield? You didn't get this problem?

RedX2501:
@larshg I pretty much downloaded what was available on the website after selecting all packages. In the csproj I fixed the paths to use `styles/github.css` instead of `src/styles/github.css`.

I was not aware of any other automated steps done to highlight.js. If it exists i totally overlooked it. Maybe we can add it to the wiki.

RedX2501:
@larshg @jakubgarfield What advantages do we have from using bower to get the highlight.js source instead of just getting the pre-compiled version from the website? Are we using any special options? How are we building highlight.js?

latop2604:
These build artefacts are really annoying.
Could we update .gitignore to hide these files?

larshg:
@latop2604 As I can figure its not build artifacts. I don't think highlight was ever builded. I'm trying to figure out how this bower thing works and it seems that it only downloads the source files, but its not compiled by bower. For that I think you need to use some grunt or the like to make this compile.

However the github.css file is referenced from the package downloaded by bower instead of the one from the prebuild library.

You can properly delete the files, but we need to update the bower.json file, so it doesn't download the old 8.0 version.

It seems cumbersome for now to get it working using bower - so I think the best solution is to remove it from bower and use the prebuild packages from their [website](https://highlightjs.org/) like @RedX2501  did.

We just need to fix a few paths, that I also mentioned earlier - any thoughts @latop2604 @RedX2501 @jakubgarfield ?

larshg:
I have created a pull request #427 to remove the old highlight version from bower.

RedX2501:
@larshg Yes, bower will only download the packages, most of them will need building. At least that's what highlight.js states.

@jakubgarfield How do you compile (if needed) font-awesome and pure?

jakubgarfield:
I'm happy for you guys to add just precompiled versions of those packages - whatever is easier to maintain and build.

jakubgarfield:
@RedX2501 I don't think I was compiling them, maybe just using them just as downloaded and put them in the normal pipeline. (but it's a quite a long time ago)

Issue #156, opened 2014-10-12
Setting username as email address resolves in invalid url for editing the user profile 


janhebnes:
You can not delete or update information on the created user when the username contains a @.

willdean:
This appears to work OK in 5.1.1 - I have created, edited and deleted a user called a@b

PR #426, opened 2016-01-23
[RFC] Add Id to Database Tables

### LINKED ISSUE #321: Adding a team with forward slashes in the names breaks access to edit/delete pages
I added a team named FB/DRF/LE and then tried to edit the team later and I get 404 errors because it throws the slashes in the URL like so: 

`http://git.abc.com/Bonobo/Team/FB/DRF/LE/Edit`

Same thing happens if I try to delete it, so now it's stuck in my system...

RedX2501:
This is an attempt to fix all the username encoding problems we have had by adding Id to the User, Repository and Team tables.
All the edit, create, etc links should now use id's instead of the names.

I have tried to make this change as shallow as possible, internally the names are still used whenever possible. This should work as i have marked the names as UNIQUE in the DB.

This patch is still missing the update procedure. which means you will have to start with a new DB to 
test.

I'm not using ADBackend if someone who is could give this a try it would be good, there are probably some bugs lurking there.
<strike>
Test wise you could try this script:

```
ALTER TABLE User RENAME TO sqlitestudio_temp_table;

CREATE TABLE User (
    Id       INTEGER       PRIMARY KEY AUTOINCREMENT
                           UNIQUE
                           NOT NULL,
    Name     VARCHAR (255) NOT NULL,
    Surname  VARCHAR (255) NOT NULL,
    Username VARCHAR (255) NOT NULL
                           UNIQUE,
    Password VARCHAR (255) NOT NULL,
    Email    VARCHAR (255) NOT NULL
);

INSERT INTO User (
                     Name,
                     Surname,
                     Username,
                     Password,
                     Email
                 )
                 SELECT Name,
                        Surname,
                        Username,
                        Password,
                        Email
                   FROM sqlitestudio_temp_table;

ALTER TABLE Team RENAME TO sqlitestudio_temp_table2;

CREATE TABLE Team (
    Id          INTEGER       PRIMARY KEY AUTOINCREMENT
                              UNIQUE
                              NOT NULL,
    Name        VARCHAR (255) NOT NULL
                              UNIQUE,
    Description VARCHAR (255)
);

INSERT INTO Team (
                     Name,
                     Description
                 )
                 SELECT Name,
                        Description
                   FROM sqlitestudio_temp_table2;
				   
ALTER TABLE Repository RENAME TO sqlitestudio_temp_table1;

CREATE TABLE Repository (
    Id            INTEGER       PRIMARY KEY AUTOINCREMENT
                                UNIQUE
                                NOT NULL,
    Name          VARCHAR (255) NOT NULL
                                UNIQUE,
    Description   VARCHAR (255),
    Anonymous     BIT           NOT NULL,
    AuditPushUser BIT           NOT NULL
                                DEFAULT ('0'),
    [Group]       VARCHAR (255) DEFAULT (NULL),
    Logo          BLOB          DEFAULT (NULL) 
);

INSERT INTO Repository (
                           Name,
                           Description,
                           Anonymous,
                           AuditPushUser,
                           [Group],
                           Logo
                       )
                       SELECT Name,
                              Description,
                              Anonymous,
                              AuditPushUser,
                              "Group",
                              Logo
                         FROM sqlitestudio_temp_table1;


ALTER TABLE UserRepository_Administrator RENAME TO sqlitestudio_temp_table0;

CREATE TABLE UserRepository_Administrator (
    User_Id       INTEGER NOT NULL,
    Repository_Id INTEGER NOT NULL,
    CONSTRAINT UNQ_UserRepository_Administrator_1 UNIQUE (
        User_Id,
        Repository_Id
    ),
    FOREIGN KEY (
        User_Id
    )
    REFERENCES User (Id),
    FOREIGN KEY (
        Repository_Id
    )
    REFERENCES Repository (Id) 
);

INSERT INTO UserRepository_Administrator (
                                             User_Id,
                                             Repository_Id
                                         )
                                         SELECT User.Id,
                                                Repository.Id
                                           FROM sqlitestudio_temp_table0
                                                JOIN
                                                User
												JOIN
												Repository
                                          WHERE sqlitestudio_temp_table0.User_Username = User.Username
										  AND sqlitestudio_temp_table0.Repository_Name = Repository.Name;

DROP TABLE sqlitestudio_temp_table0;

ALTER TABLE UserRepository_Permission RENAME TO sqlitestudio_temp_table0;

CREATE TABLE UserRepository_Permission (
    User_Id       INTEGER NOT NULL,
    Repository_Id INTEGER NOT NULL,
    CONSTRAINT UNQ_UserRepository_Permission_1 UNIQUE (
        User_Id,
        Repository_Id
    ),
    FOREIGN KEY (
        User_Id
    )
    REFERENCES User (Id),
    FOREIGN KEY (
        Repository_Id
    )
    REFERENCES Repository (Id) 
);

INSERT INTO UserRepository_Permission (
                                          User_Id,
                                          Repository_Id
                                      )
                                      SELECT User.Id,
                                             Repository.Id
                                        FROM sqlitestudio_temp_table0
                                             JOIN
                                             User
											 Join
											 Repository
                                       WHERE sqlitestudio_temp_table0.User_Username = User.Username
									   AND sqlitestudio_temp_table0.Repository_Name = Repository.Name;

DROP TABLE sqlitestudio_temp_table0;

ALTER TABLE UserTeam_Member RENAME TO sqlitestudio_temp_table0;

CREATE TABLE UserTeam_Member (
    User_Id INTEGER NOT NULL,
    Team_Id INTEGER NOT NULL,
    CONSTRAINT UNQ_UserTeam_Member_1 UNIQUE (
        User_Id,
        Team_Id
    ),
    FOREIGN KEY (
        User_Id
    )
    REFERENCES User (Id),
    FOREIGN KEY (
        Team_Id
    )
    REFERENCES Team (Id) 
);

INSERT INTO UserTeam_Member (
                                User_Id,
                                Team_Id
                            )
                            SELECT User.Id,
                                   Team.Id
                              FROM sqlitestudio_temp_table0
                                   JOIN
                                   User
                                   JOIN
                                   Team
                             WHERE sqlitestudio_temp_table0.User_Username = User.Username AND 
                                   sqlitestudio_temp_table0.Team_Name = Team.Name;

DROP TABLE sqlitestudio_temp_table0;

ALTER TABLE UserRole_InRole RENAME TO sqlitestudio_temp_table0;

CREATE TABLE UserRole_InRole (
    User_Id   INTEGER       NOT NULL,
    Role_Name VARCHAR (255) NOT NULL,
    CONSTRAINT UNQ_UserRole_InRole_1 UNIQUE (
        User_Id,
        Role_Name
    ),
    FOREIGN KEY (
        User_Id
    )
    REFERENCES User (Id),
    FOREIGN KEY (
        Role_Name
    )
    REFERENCES Role (Name) 
);

INSERT INTO UserRole_InRole (
                                User_Id,
                                Role_Name
                            )
                            SELECT User.Id,
                                   Role_Name
                              FROM sqlitestudio_temp_table0
                                   JOIN
                                   User
                             WHERE sqlitestudio_temp_table0.User_Username = User.Username;

DROP TABLE sqlitestudio_temp_table0;

ALTER TABLE TeamRepository_Permission RENAME TO sqlitestudio_temp_table0;

CREATE TABLE TeamRepository_Permission (
    Team_Id       INTEGER NOT NULL,
    Repository_Id INTEGER NOT NULL,
    CONSTRAINT UNQ_TeamRepository_Permission_1 UNIQUE (
        Team_Id,
        Repository_Id
    ),
    FOREIGN KEY (
        Team_Id
    )
    REFERENCES Team (Id),
    FOREIGN KEY (
        Repository_Id
    )
    REFERENCES Repository (Id) 
);

INSERT INTO TeamRepository_Permission (
                                          Team_Id,
                                          Repository_Id
                                      )
                                      SELECT Team.Id,
                                             Repository.Id
                                        FROM sqlitestudio_temp_table0
                                             JOIN
                                             Team
											 JOIN
											 Repository
                                       WHERE sqlitestudio_temp_table0.Team_Name = Team.Name
									   AND sqlitestudio_temp_table0.Repository_Name = Repository.Name;

DROP TABLE sqlitestudio_temp_table0;
DROP TABLE sqlitestudio_temp_table;
DROP TABLE sqlitestudio_temp_table1;
DROP TABLE sqlitestudio_temp_table2;

```
</strike>
Fix #321, #156


RedX2501:
Also a test is failing because the password Service Provider will want to Update the password if it is an old one. I don't know how to mock the BonoboContextProvider to fix the test.

Further all links to user profiles and repos in team or anywhere else is broken because i have not yet figured out how to properly pass Models to the views.

Any help solving these problems is very welcome.

Bonobo.Git.Server/Data/Update/InsertDefaultData.cs
```diff
@@ -14,8 +14,8 @@ public string Command
                 return @"
 
                     INSERT INTO [Role] ([Name], [Description]) VALUES ('Administrator','System administrator');
-                    INSERT INTO [User] ([Name], [Surname], [Username], [Password], [Email]) VALUES ('admin', '', 'admin', '0CC52C6751CC92916C138D8D714F003486BF8516933815DFC11D6C3E36894BFA044F97651E1F3EEBA26CDA928FB32DE0869F6ACFB787D5A33DACBA76D34473A3', '');
-                    INSERT INTO [UserRole_InRole] ([User_Username], [Role_Name]) VALUES ('admin', 'Administrator');
+                    INSERT INTO [User] ([Name], [Surname], [Username], [Password], [Email]) VALUES ('admin', '', 'admin', '28493778DD028138E5F43E32F4CF2111B632BCF5AC4BFE8BCDFCAF77696ACE04B3ABDDF8B977A2D7A0FBDEC2678AB5E392259986C1D08498F1C6FD287BE7706A', '');
```
latop2604: Why did you update admin password hash?

latop2604:
I think it would be preferable to add a new IUpdateScript class and append it to UpdateScriptRepository.
So the update process will be fine

latop2604:
I can create the T-sql version if you want.

Bonobo.Git.Server/Data/Update/InsertDefaultData.cs
RedX2501: I had to Update the hash because now it's computed with the user id instead of username.this means we could implement username change in the future without having to update the hash.

latop2604:
About the CorrectMd5PasswordsWillBeUpgraded test, you need to extract your code to a lambda like `_updateUserPasswordHook` and put the query in `EFMembershipService`.

```c#
Func<string, int?> findUserIdFromUsername => (username) => 
{
    var user = db.Users.FirstOrDefault(i => i.Username == username);
    if (user != null)
    { 
        return user.Id;
    }
    return null;
}
```

Ideally the two lambda would be converted to an interface.
(Sry for my english)

RedX2501:
I noticed that changing the way the password is generated without changing its format is not a good idea. You cannot differentiate between them and i will be undoing that change.

RedX2501:
So this version now should have everything working except the upgrade path. I still need to add that.

RedX2501:
This version also Requires MVCCheckBoxListFor. I haven't yet added it properly to the project because in my hotel room nuget is refusing to work.

larshg:
I could only find MvcCheckBoxList not MVCCheckBoxListFor under nuget - might be why its not working? :) Anyways - it seems to work with cookie auth :) Just a minor merge conflict with the PR I made #429  - I'll try out AD soon :)

RedX2501:
@larshg  thank for trying it out.

Yeah it's http://mvccbl.com/

    Install-Package MvcCheckBoxList

larshg:
Its really troublesome to get ids added to the users found through AD. I think I got something working - but if you delete all files in the ADBackend/Users and there has been added a new user meanwhile, the Ids get lost as they will be saved in the files. Is this acceptable?

There was a few places that needed some corrections - I'll try open a PR to your branch @RedX2501.

Also for some reason the MembershipService is not instantiated when running the UpdateTeams() so it fails on Null pointer exception when it tries to get the Users.

Do you have an idea why its not instantiated :) ?

larshg:
```       public static int Id(this IPrincipal user)
        {
            int ret = 0;
            int.TryParse(user.GetClaim(ClaimTypes.Upn), out ret);
            return ret;
        }```

This doesn't work with AD users as this would be an email string.

Maybe we could use a GUID instead? An AD User has such and ID at least.

RedX2501:
@larshg we could try to convert the username + domain to a numeric value and use that as an id. 

The problem with guid is that they have letters don't they? If they do not, use the guid. 

There is no max length requirement for the id anywhere.

larshg:
Just out of curiosity why do we have to change from string to int - and what coudn't handle \\ in usernames?

RedX2501:
@larshg Urls do not like / and % in them so there was no way of having the domain with the username in the current way.  I thought that replacing it with a numeric id would be easy. I guess i was wrong.

larshg:
And this will happen if you use windows authentication and internal database for membership service? Because at my setup using windows auth and AD membership service its using the mail as identifier and here @ gets changed to %40 in the URL but it works :)

Well, we just need to figure out how to generate an integer from either the UPN - but this can be anything from a mailadress to something more exotic I guess :) (I'm a self learned admin on a single server at my company - so I'm not that familiar with all this :D )

larshg:
Perhaps we can change the URL path and routing like they suggest in [this](http://stackoverflow.com/questions/5901049/how-to-url-encode-parameters-in-asp-net-mvc) thread. Using a querystring to avoid the username being part of the URL :) ? @jakubgarfield what do you think?

larshg:
Just did a quick test and it seem to work fine. If we encode usernames with HttpUtility.urlencode and change the routing to use querystring it allows the escaped version of \\ to be %5C etc.

RedX2501:
Somehow  I had the impression that % were not allowed in  URL according to https://tools.ietf.org/html/rfc3986 but what really is not allowed are single % signs.

On the long run the advantages of an id instead of a name is that you can rename users/groups/teams/repos without links becoming invalid and you needing to update all the related tables in the DB..

My goal with this patch was to internally keep using `Username` for as much as possible, that's why i made the table fileds unique.

I thnk the changes to `User.Id` was not necessary and I might be able to roll it back. I'd still rename it to `User.Uid` to make it clear it is not related to the database Id field.

Also I wanted to be able to display a Nice Name and User id on all the fields that link to the user because in our company the userids are nonsense and seeing a bunch of xy2z7h83 in the user list makes it hard to find the person you are looking for. This might have been better left to another patch.

larshg:
Just for info - I'm still working on using GUID instead of integers - as the AD Users/groups has a GUID already and hence its easier than having to generate integers for AD Users that might be added or removed and therefore complicate things when generating the integer ids.

I'll try make PR to this soonish, but I'm a bit overloaded at work :S

jakubgarfield:
Don't worry about that and take all the time you need :)

RedX2501:
@larshg Thanks for the guid changes, i guess now we are at the home stretch. All i need to figure out is how to write the update script.

RedX2501:
@larshg  @willdean  @jakubgarfield 
This should be the last commit needed for this patch.

I'd ask you to please test this with your full dbs, just copy the db files. The repos do not need to exist.

Make sure you test all the functionalities as pretty much everything was touched.

I had a merge confilct on the Password reset dialog. So please test that too as i don't have a working mail server here.

willdean:
This is good timing, because I've written a whole load of unit tests for the EFMembershipService, and it's amusing to see how that do with the new code...

So far:

* When it's a function argument in EFMemberShipservice, the new 'Id' is sometimes called 'id' and sometimes called 'guid' - I think we should standardise (probably on 'id').

* As far as I can tell, AutomaticUpdater.DoUpdate is all EF-based and could run on either Sqlite or SqlServer (obviously it has different scripts)   The new AddGuidColumn() is hardcoded to assume SQLite.

* I have not been following along here, but why can we not do an ALTER TABLE to add the new column to the existing tables?  

Anyway, I'm going to try and make it work with my tests, and let you know what broke

willdean:
AddGuidColumn.CreateTables contains code for building all the new tables - which must duplicate the code InitialCreateScript.

This feels like there's a risk that they'll diverge - could AddGuidColumn not call the InitialCreateScript?

RedX2501:
@willdean probably i can call  the create table script after i have renamed the old tables, I'll look into it.

GUID is a primary key and a such must be unique. Since it is a char there is no way of providing a useful default value for it. I could not find anything if functions are allowed as default values.

SQLite also does not offer a create GUID function, i could create one but since i was doing the update for the user i did it for the rest in the same place.

RedX2501:
@willdean are those unit tests or integration tests? What framework did you choose?

willdean:
> GUID is a primary key and a such must be unique. Since it is a char there is no way of providing a useful default value for it. I could not find anything if functions are allowed as default values.

That makes sense - I realised after I'd said it that an ALTER might be really hard because of the PK stuff.

My tests are just part of the test DLL, so they use MSTest - if you like, I'll commit them onto your branch once I have them working again.   They have found a couple of bugs in the new EFMembershipService...  :-)

larshg:
That would be awesome - I'll also really like to see the test etc :) There is however still some missing changes about the guid that needs to be merged into this one :D - but hopefully I get the PR cleaned for @RedX2501

willdean:
OK, there seems to be quite a serious problem with password hashing - some parts of the code are using the ID guid as the salt for the password, and some parts are using the username.

IMO, using the username is no good, because if you changed your username, then your password would become invalid.

To avoid this confusion, GetSaltedHash should probably take a Guid rather than a string, then it would be impossible to get this wrong.

If you're OK with this, I'm most of the way through fixing this, but I wanted to check what the intent was here.

Really I don't think we should use any part of the user profile as a salt, we should just use another random number, but then we'd have to store that too.

Is there any good reason not to use the GUID as the salt?

willdean:
OK, last words on this from me now - I've got it all working in my tests with a fresh database.  I have messed around trying to make the migration stuff work in the test harness, but without complete success.

If there's another big patch coming from @larshg then I think I'll wait until that's in to look at it again.

It's a fantastic bit of work and it's going to allow about half the open issues to be be fixed...   But it's not quite ready yet IMO.

RedX2501:
@willdean One thing about the passwords. If you don't change the format you cannot change how they are generated because you can no longer differentiate between new and old passwords. This patch at one point tried to update the password function to use the Id as salt but I rolled those back as soon as i realized the problem.

willdean:
@RedX2501 Yes, sorry, my mistake - I had thought the change in hash format was part of this change, but it's earlier.  I have had a different idea about dealing with the salt problem, but I will put it in another issue, so it doesn't get lost in here

RedX2501:
I added the  latest PRs from will and lars to this aswell as trying an SqlServer version of the AddGuidColunm. If anyone could test that it works that would be nice as i don't have yet a working SQLServer and i'd really like to save myself the hassle.

willdean:
I am (somewhat unsuccessfully right now ) trying to expand my EF test setup to use a LocalDB instance of SqlServer to test this latest stuff - probably won't be any good for Appveyor, but might be useful to individual devs.  Will PR this when/if it's working

willdean:
I think SqlServer has much more capable ALTER abilities, so we may be able to just add the column, fill it and then set it to be PK.  I have a LocalDB instance running under nCrunch now, so it's quite easy for me to experiment with this

willdean:
@RedX2501 I've sent you a PR which contains localdb-based automated test which uses SQL Server

I made a bunch of changes to the SQL in AddGuidColumn to get it to run, but I have stuck with your approach of making new tables and copying everything.

I played with doing alter-column stuff, but by the time you've finished dropping and reconstructing all the foreign-key constraints it's not going to be any less code really.

The two AddGuidColumn implementations are very similar and could possibly be combined - not sure if that's worth it.

'user' is a keyword in T-SQL, so I had to bracket that everywhere.

The SQL server tests are rather slow to be called 'unit tests', but they do run OK on AppVeyor / nCrunch, so I didn't label them 'integration tests'.

RedX2501:
Ok it seems to me that the PR and additions that this PR is starting to receive are no longer directly related to the ID functionality.

I would also say that we should have tested this enough then since noone seems to have found anything that was not working for some time.

So I'd propose that we merge this and then start doing small PRs against master.

If noone has anything agains it i'll merge this on sunday night (indian time, that should be around lunch in the US).

willdean:
:+1: on the merge.

I'd like to keep adding the account test stuff, because it keeps finding little bits and pieces some of which are to do with the GUID port.

Do you think it would be a good idea for the Sqlite AddGuidColumn to take a backup of the .db file before it starts?  I know it all runs in a transaction, but just in case we trash someone's database or they want to revert to the previous version of Bonobo?

RedX2501:
That is a tricky question. Do it once and the users will expect that to
happen every time.

Since the documentation clearly states to make a backup before updating I'd
say no, we don't make a backup.

willdean:
Good point.  Maybe we just need to have it in double-height bold in the release notes for this next release.

willdean:
@RedX2501 - Can you check if the user-name checkboxes on the repo edit page  (Repository/Edit.cshtml) are all screwed-up (you need to have several users to be able to see it)  ?  Something seems to be confusing the CheckBoxListForCall, but there have been a lot of changes there and I'm not sure what's what.

larshg:
@willdean they are are not correct. They seem to be appended but not with a line break, so its really confusing to select the right ones :-) I think de starter using listboxfor lib from nuget? So probably a setting is missing...

RedX2501:
The Boxes work only the layout is screwed. I'll still merge and later post
a patch for the layout.

I'd like to get integration test working soon.
Am 21.02.2016 11:46 schrieb "Lars" <notifications@github.com>:

larshg:
I had to change the sqldb connection string since I have localdb (v.12) 2014 installed - but else it finished all tests. There is however 13 skipped tests? - also all the localdb tests takes about 500ms but I guess it restarts the localdb instance each time? - or why the extra time compared to sqlite which are under 10ms most of the time.

willdean:
@larshg Yes, the SqlServer tests are very slow because it's using a file-based database and creating it from scratch each time.  The Sqlite tests are using an in-memory database which is hugely quicker.

I think the SqlServer stuff could be much faster if you could reliably drop the database but keep the same file - there are lots of article around about how to do this, but I couldn't get it working reliably myself.

Not sure what the best thing to do about the v11/v12 thing - I have both here but probably one is from VS2013 and one from VS2015.  I think the \v11.0 in the connection string is really an instance name rather than a version number, but I'm unclear how the various instances get created.

What does "sqllocaldb info" at a command prompt show?

willdean:
@larshg The skipped tests are mostly ADTests, which I couldn't get working because I don't have a working domain server.   If you liked to have a look at ADMembershipServiceTest and see if you can make them work then that would be cool.

There are another couple of tests with Ignores on them - I try to comment why I've done it for each one.   A global search for `ignore]` will find them

RedX2501:
I'd like to summarize all the chanegs we did for the release notes.

Functional changes:
- Added IDs to User, Repository, Team, Role tables. This allows implementation of rename features in the future.
- Added human friendly names to whenever you can select users.
- Repository names are no longer case sensitive

Please ping me if i forgot something.

PR #454, opened 2016-02-13
Added a button to rescan the repo directory

willdean:
This adds a button to the repository index which re-runs the directory scan, that's normally only run at startup.  This avoids the need to restart the whole app (which tends to need access to the server) if you've just added a repo directly to the file system.

* I passed a dummy Id to the action method to avoid needing to mess with the routing table.   
* The messing about with the form for the button is so that the action is POSTed rather than GET.

### DIFF
--- a/Bonobo.Git.Server/Content/Site.css
+++ b/Bonobo.Git.Server/Content/Site.css
@@ -108,6 +108,7 @@ nav.branches ul li:hover ul li { float: none; }
 .Tree #files .date { width: 150px;}
 .breadcrumbs { padding: 7px; margin: 0; }
 .breadcrumbs em { font-weight: bold; font-style: normal; }
+.inline-form { display: inline-block; }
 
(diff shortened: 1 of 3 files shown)


RedX2501:
I'm not sure how i feel about every user being able to do this. Did you have a special use-case in mind when you made it public?

Else it LGTM.

willdean:
The intention was admin-only - I think I misread the @if in the view and didn't add any server-side validation either.   I'll improve that.

Like you, I don't see a legitimate non-admin use.

RedX2501:
I made a mistake and only tested with the admin account.

If you set the option 'users can create repos' then rescan also appears. Not sure if that is your intention.

willdean:
It wasn't my intention - I didn't read the view code properly.  I'll fix it later.

willdean:
OK, all fixed, definitely admin-only now  - I squashed/rebased this to be one commit

PR #460, opened 2016-02-15
[RFC] Allow push-to-create

### LINKED ISSUE #376: Scripted Repository Creation
It would be wonderful if we could create repositories from command line in order to automate some processes !!

willdean:
Resolves #376.  I made it an option in user config, so it's not available by default.  At present, only Admins can push a new repo.  

I don't really know if this is a good idea or not, but it does allow automated creation of repos from the client, which is apparently what the OP on #376 wanted.

### DIFF
--- a/Bonobo.Git.Server/App_GlobalResources/Resources.Designer.cs
+++ b/Bonobo.Git.Server/App_GlobalResources/Resources.Designer.cs
@@ -1827,6 +1827,24 @@ namespace Bonobo.Git.Server.App_GlobalResources {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Allow push to create repositories.
+        /// </summary>
+        public static string Settings_Global_AllowPushToCreate {
+            get {
+                return ResourceManager.GetString("Settings_Global_AllowPushToCreate", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to If checkbox is checked, Administrators can push a new repository .
+        /// </summary>
+        public static string Settings_Global_AllowPushToCreate_Hint {
+            get {
+                return ResourceManager.GetString("Settings_Global_AllowPushToCreate_Hint", resourceCulture);
+            }
+        }
+        
+        /// <summary>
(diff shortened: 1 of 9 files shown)


RedX2501:
Thanks for the patch, i'll look into it. I just wanted to tell you that if instead of "addresses" you use one of the keywords [here](https://help.github.com/articles/closing-issues-via-commit-messages/) the PR and the issue are automatically linked and close when the PR is merged.

willdean:
Cool!  I didn't know that, thanks.

RedX2501:
LGTM (without testing).

Maybe we should add also the option to allow everybody push to create. Maybe a radio button with

Push to create:
- [ ] Off
- [ ] Admin
- [ ] All

option.

willdean:
Could that perhaps use the existing 'allow users to create repositories' option?

RedX2501:
I think so, please expand on the tooltip of the 'allow users to create repositories' option to indicate this side-effect if 'allow push to create' repos is on.

I gave it a quick spin, nice job.

I think tho that the user that created the repo should be automatically added as an administrator of the repo.

jakubgarfield:
Nice feature! Thanks for that @willdean :+1:

willdean:
I'm closing this for now because I think the auth is wrong and we shouldn't merge it yet.

RedX2501:
What makes you think that the auth is wrong?

willdean:
It would allow creation with anon push, which I don't think was my intent, and I wanted to add support for 'allow users to create repos' which caused me a crisis of confidence about the whole thing.  

Separately I think that RepositoryPermissionService.HasPermission is rather over-worked and under-named in this context - I'm (mis)using it to detect that a user is an admin, which isn't really what it's intended for anyway.  I may put it back to returning False for any missing repo (which is how it used to work) and detect 'permission to create' separately. 

I think the contract for IRepositoryPermissionService.HasPermission needs clearer definition anyway - there are subtle differences in behaviour between the AD and EF implementations which are bound to cause confusion at some point.

If I could face it, I'd try and built some more testing, because that's at least an unambiguous test of a contract.  I don't know how to test the AD stuff without a domain server though, which is a huge pain.

PR #464, opened 2016-02-18
Adding an integration test framework.

RedX2501:
Unit tests only cover a small part of the functionality of complex web applications.
For instance it's not possible to validate the whole chain from password input to redirect with such tests.

To cover these use cases I will try to provide a base for adding automated Integration tests into the repo.

I have chosen [SpecsFor.Mvc](http://specsfor.com/SpecsForMvc/default.cshtml) as the framework as there are not many and this one still seems to be supported. Please don't be shy if you have a better alternative. All criticism are welcome.

I need feedback on wether i should keep it as a separated project or merge into the Tests project and add the `[Integration]` attribute.

I'm having an issue, that SUT is currently null and I don't know why, i have asked (https://github.com/MattHoneycutt/SpecsFor/issues/86) about it and am awaiting a reply.

Although we will only be able to test CookieAuth with InternalDb at the beginning i think this is an important step to have less regression failures in the future.


willdean:
I don't have any useful opinion about the specific tool to use, but I'm +100 on improvements to testing, code-quality, etc.

Sounds like a really good idea.

Personally I don't think having it in the main test DLL is a big problem for now - there's already stuff in there which is disabled for AppVeyor, so having a bit more is no problem.  Managing extra projects is all overhead.

RedX2501:
@jakubgarfield @willdean @larshg 
The second commit adds a script to automatically fetch git 2.6.1 from github. Unfortunately the exe does not provide any means to extract itself via command line. SO 7z is needed.

Should I commit a version into the repo? Should we just warn the user that it needs 7z.exe in the PATH?

willdean:
Why not just commit the git parts you need - doesn't seem much point committing an exe to unpack another file just so that you then don't have to commit the other file.

RedX2501:
@willdean The idea was to allow for different git versions when testing. Git also heavily relies on the whole msys environment for tools like diff, patch, etc.

jakubgarfield:
The initial tests were a bit hacky and not very CI friendly so I really like where this is going :+1:

RedX2501:
So it seems that everything is working now except one test. Thats the Command line test. It works locally if I run run-tests.bat.

The odd part is that the linenr in the failed test does not match the source code. Anyone got any idea why that could be?

RedX2501:
So all is working. If there are no objections i'll merge it by tomorrow night (indian time).

RedX2501:
@willdean It seems that AppVeyor does not like your tests. Have any idea why? They run fine on my computer.

RedX2501:
Ok so it was my mistake, it all seems to be fine now.

willdean:
Awesome :+1:

Issue #459, opened 2016-02-15
AD authentication not working for Administrator access


SGatOCCU:
Hello All,

I'm having an issue getting an administrative login via Active Directory.  I am able to successfully login using AD, just not as an administrative user.

Unexpected behavior:
- When attempting to login as an administrative user I can login, but receive the error "Unauthorized
You don't have permission for the selected action or page."

Troubleshooting:
- The machine is connected to AD.  I can successfully login with an AD user as a regular user.
- Nested groups are not being used.
- Case sensitivity is not the issue - I've tried case correct and lower case.
- Local admin will not work when AD authentication is used.  This is to be expected I suppose, but limits workarounds.

Here is the user authentication block from the web.config file:
```
  <add key="UserConfiguration" value="~\App_Data\config.xml" />
    <add key="DefaultRepositoriesDirectory" value="~\App_Data\Repositories" />
    <add key="GitPath" value="~\App_Data\Git\git.exe" />
    <add key="GitHomePath" value="~\App_Data" />
    <add key="RecoveryDataPath" value="~\App_Data\Recovery" />
    <add key="AuthenticationProvider" value="Cookies" />
    <!--<add key="AuthenticationProvider" value="Windows" />-->
    <!--<add key="AuthenticationProvider" value="Federation" />-->

    <!--<add key="MembershipService" value="Internal" />-->

    <add key="MembershipService" value="ActiveDirectory" />
    <add key="ActiveDirectoryDefaultDomain" value="ulaneocu.org" />
    <add key="ActiveDirectoryBackendPath" value="~\App_Data\ADBackend" />
    <add key="ActiveDirectoryMemberGroupName" value="Git1101Gitall" />
    <add key="ActiveDirectoryRoleMapping" value="Administrator=Git1101Admin" />
    <!--<add key="ActiveDirectoryTeamMapping" value="Developers=GitTeam" />-->

```
Thanks in advance for your assistance and time,
Sean

willdean:
I may well reveal my almost complete ignorance about AD here, but does your user show any roles if you go to the users page?  Seems to me you'd be unlikely to fail any auth tests if you correctly have the role 'Administrator'

SGatOCCU:
Hi Will.  Thanks for the reply.

[...]

PR #468, opened 2016-02-21
Add Diagnostics page

willdean:
This PR adds a page /Home/Diagnostics (which is only available when run locally).

This contains a report of lots of details about the configuration (both WebConfig and UserConfig), plus the last 10K of the log file.

The idea would be that this could be a help when people report issues, many of which are configuration based or can be diagnosed by reference to the log.

Obviously dumping chunks of config is a potential security issue, but I feel this is suitably mitigated by the IsLocal check - by intention we DON'T want to use any kind of authentication/authorization, because we might be trying to diagnose problems with them.

An alternative to actually just returning the data as the page would be to write the report to somewhere on the machine (app_data probably), which would doubly-verify that someone had local access to the server.

Thoughts?

### DIFF
--- a/Bonobo.Git.Server/Bonobo.Git.Server.csproj
+++ b/Bonobo.Git.Server/Bonobo.Git.Server.csproj
@@ -267,6 +267,7 @@
     <Compile Include="Attributes\RepositoryNameNormalizerAttribute.cs" />
     <Compile Include="Configuration\ActiveDirectorySettings.cs" />
     <Compile Include="Configuration\AuthenticationSettings.cs" />
+    <Compile Include="Configuration\DiagnosticReporter.cs" />
     <Compile Include="Configuration\FederationSettings.cs" />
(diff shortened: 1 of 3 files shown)


RedX2501:
I had a look at it and it seems really nice and helpful.

I dislike though that you dump all usernames. If you are using AD or WindowsAuth that is not so nice. Maybe we should only dump the number of users and admins?

Another thing is the readable checks, they should IMHO be writability check, since that is what the user has to enable. Not sure how to test that though.

Interesting would also be the kind of authentication chosen: Windows or Cookie.

willdean:
OK on suppressing names - I guess that's a reasonable compromise - I added that because I was trying to help the poster in #459 and thought there was some name/group issue.  But he promptly elided them all from the results anyway, so they weren't any use.

If people feel the need to hide them from the lists, then there's no point including them - the count's a good idea, and I can add the rest too.

RedX2501:
LGTM

PR #474, opened 2016-02-22
Fix for ADRoleProvider

### LINKED ISSUE #459: AD authentication not working for Administrator access

chrismdev:
Hopefully this fixes #459

### DIFF
--- a/Bonobo.Git.Server/Security/ADRoleProvider.cs
+++ b/Bonobo.Git.Server/Security/ADRoleProvider.cs
@@ -45,7 +45,8 @@ namespace Bonobo.Git.Server.Security
 
         public string[] GetRolesForUser(Guid userId)
         {
-            return ADBackend.Instance.Roles.Where(x => x.Members.Contains(userId.ToString(), StringComparer.OrdinalIgnoreCase)).Select(x => x.Name).ToArray();
+            var user = ADBackend.Instance.Users.Where(x => x.Id.Equals(userId)).FirstOrDefault();
+            return ADBackend.Instance.Roles.Where(x => x.Members.Contains(user.Username, StringComparer.OrdinalIgnoreCase)).Select(x => x.Name).ToArray();
         }
 
         public string[] GetUsersInRole(string roleName)


willdean:
This is clearly a fix for a bug, but sadly I doubt it's a fix for  #459 problem, because the bug you're fixing is a regression on the Guid ID work, which wasn't in the version which is being run in #459

chrismdev:
Fair enough. It does fix the same issue in master though.

willdean:
ok, cool, TVM.   I've merged it, though I will probably make a minor stylistic tweak in that the rest of the project compares GUIDs with == rather than .Equals.    

I think really RoleProvider has far to much username stuff going on in it - it should all be guids now, but anyway...

xiaoyuvax:
i still meet problem #459 with the officially downloaded binary,when there will be a update?

willdean:
@xiaoyuvax I don't think there's a problem as general as "administrator role doesn't work for AD" - it's more likely something about your particular setup which Bonobo doesn't work well with, and hopefully you'll be able to help us work out what it is.

Could you please have a look in the app_data\logs directory for the latest log, and see if you can see in there anything which gives a clue what might be happening - feel free to attached it here if you want me to have a look.

xiaoyuvax:
@willdean 
the log attached for you analysis (i guess that a few chinese characters won't frustrate you ;-):

[log-20170624.txt](https://github.com/jakubgarfield/Bonobo-Git-Server/files/1099710/log-20170624.txt)

willdean:
Thanks for that - would you mind looking in web.config and changing this line:

```
<add key="serilog:minimum-level" value="Information" />
```
 to say 

```
<add key="serilog:minimum-level" value="Verbose" />
```

Then try again and give us the logs again.   For some reason I'm not quite sure about it's losing the domain name somewhere.

xiaoyuvax:
[log-20170624_new.txt](https://github.com/jakubgarfield/Bonobo-Git-Server/files/1099845/log-20170624_new.txt)

note that server fs. is down.

willdean:
@xiaoyuvax - thanks for the logs - can we move to #706 please

Issue #475, opened 2016-02-23
RepositoryPermissionService User.Id is null


chrismdev:
Hi,
When trying to clone a repo I am getting either authentication failed or 500 errors because the User.Id is null in the GitController, specifically this line:

bool hasPermission = RepositoryPermissionService.HasPermission(User.Id(), repositoryName);

I'm using windows authentication with the active directory membership service. The user I am testing with is an administrator (ie: is a member of this group: ="ActiveDirectoryRoleMapping" value="Administrator=GitAdmins" ).

I can clone the repo if I set the anonymous flag.

chrismdev:
This is my workaround for the problem. Is this safe?

   public ActionResult SecureGetInfoRefs(String repositoryName, String service)
        {
            if (!RepositoryIsValid(repositoryName))
            {
                return new HttpNotFoundResult();
            }

            bool hasPermission = false;
            string username = string.Empty;
            string password = string.Empty;
            if (User.Id() == Guid.Empty)
            {
                var basicAuth = Request.Headers["Authorization"];
                if (!string.IsNullOrEmpty(basicAuth))
                {
                    var base64 = basicAuth.Split(' ');
                    var plainTextBytes = System.Convert.FromBase64String(base64[1]);
                    var usernameAndPassword = System.Text.Encoding.UTF8.GetString(plainTextBytes);

                    if (!string.IsNullOrEmpty(usernameAndPassword))
                    {
                        var usernameAndPasswordArray = usernameAndPassword.Split(':');
                        username = usernameAndPasswordArray[0];
                        password = usernameAndPasswordArray[1];
                        System.Diagnostics.Debug.WriteLine("username : " + username);
                    }
                }
                hasPermission = RepositoryPermissionService.HasPermission(username, password, repositoryName);
                System.Diagnostics.Debug.WriteLine("hasPermission: " + hasPermission);
            }
            else
            {
                hasPermission = RepositoryPermissionService.HasPermission(User.Id(), repositoryName);
            }

            bool allowAnonClone = RepositoryPermissionService.AllowsAnonymous(repositoryName);
            bool isClone = String.Equals("git-upload-pack", service, StringComparison.OrdinalIgnoreCase);
            bool isPush = String.Equals("git-receive-pack", service, StringComparison.OrdinalIgnoreCase);
            bool allowAnonPush = UserConfiguration.Current.AllowAnonymousPush;

[...]

PR #476, opened 2016-02-23
Attempt at a fix for #475

chrismdev:
Attempt at a fix for #475. Apologies for my coding style.

### DIFF
--- a/Bonobo.Git.Server/Controllers/GitController.cs
+++ b/Bonobo.Git.Server/Controllers/GitController.cs
@@ -19,7 +19,7 @@ namespace Bonobo.Git.Server.Controllers
     {
         [Dependency]
         public IRepositoryPermissionService RepositoryPermissionService { get; set; }
-        
+
         [Dependency]
         public IGitService GitService { get; set; }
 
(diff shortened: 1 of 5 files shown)


RedX2501:
I don't think that is the right approach as the User.Id should NEVER be empty if you are authorized.

My guess is that due to the background service in AD that fills in the users it has not been added yet.

@larshg Maybe we need to revert https://github.com/jakubgarfield/Bonobo-Git-Server/commit/2af315756ebec6bfcf05f43b09d3fd073c186186?

chrismdev:
But aren't we talking about 2 different contexts? The browser context and the git context. Yes the User.Id should never be null from a browser but git does not have access to that cookie/session. Git has to re-authenticate somehow.

RedX2501:
Git does a full http authentication (cookie or windows auth) before it reaches the GetSecureInfoRefs part.

Set a breakpoint at the authentication code and you will see it.

larshg:
@chrismdev you can try debug through the OnAuthorization in GitAuthorizeAttribute - this should authenticate you when using a git client.

And the git client should prompt for username and password if you use the general url - also if you use credential store you might have to clear it, if you have old values saved etc.

RedX2501:
@larshg the problem is that we never get to see the passwords unless internal auth is used. IIRC This means the pipeline will auth the user against the AD and let it proceed. The next step will then try to fetch this user again and becaus it has not been added to the backend by the background worker it will still have an GUID of 0.

larshg:
@RedX2501  what do mean that we never get to see the passwords? Your git client has to be authenticated somehow by providing a username and password?

RedX2501:
You are right on the git side. I was thinking about the Browser side when using Windows Auth.

larshg:
So if you run your site as a intranet and you should automatically be authenticated, bonobo is not presented with username and password?

Because I still have to supply user and password (atleast first time when running chrome) - microsoft edge browser ask for it each time its started - but maybe the url of the remote and local bonobo server I have, is not correctly added as a trusted site :)

RedX2501:
Yes, the first time you set username and password (or not if internet explorer does the ntlm handshake) and then a cookie is set on your system. Based on that cookie the user is then retrieved from the DB.

chrismdev:
Have created a new pull request for this

Issue #471, opened 2016-02-21
Upgrade to guid causes problems for people with the Upn claim in their cookies


willdean:
We need to somehow flush people's cookies when they upgrade to the GUID version, because otherwise they may have a Upn claim which is not a GUID, it's a username.  That then causes a problem in UserExtensions when it tries to parse into a GUID.

See #407

willdean:
Or perhaps we just need to change UserExtensions.Id to use TryParse

RedX2501:
I guess we can check it the beginning of authentication if the Upn contains a valid GUID. If not reset the cookie and forward to login.

Here is how to remove the cookie:
http://stackoverflow.com/questions/5122404/how-do-you-clear-cookies-using-asp-net-mvc-3-and-c/5122611#5122611

willdean:
We probably don't need to remove the cookie ourselves - if we just refuse to Authenticate at that point, then I suspect the framework will deal with it - it will do a redirect-to-login, which will then give us a decent cookie.

willdean:
I'm closing this, because I think #488 fixes/avoids it

PR #477, opened 2016-02-23
Possible fix for #475, #471

willdean:


Bonobo.Git.Server/Extensions/UserExtensions.cs
```diff
@@ -32,7 +32,13 @@ public static string GetClaim(this IPrincipal user, string claimName)
         public static Guid Id(this IPrincipal user)
         {
             string id = user.GetClaim(ClaimTypes.Upn);
-            return id != null ? Guid.Parse(id) : Guid.Empty;
+            if (id == null)
+            {
+                return Guid.Empty;
```
RedX2501: As you said having Guid.Empty is maybe not the best idea. Why not have this be a `Guid?` and keep returning null?

willdean: You are quite right to pull this up.

Yes, I am really unsure about this, and I think we should really make an effort now to decide on the rules and then be consistent about them.

I don't think there's any moral difference between `Guid.Empty` and `Guid?==null` - they're both reporting an error condition in-line with the data, and they'll both spread like a cancer across all the code, where every Guid will either be nullable or can potentially be Guid.Empty.

It's because we're (I'm) abusing this 'Id()' method to do more than one thing.

There are three separate activities here:

* Does the ClaimsUser have an ID?
* Does the ClaimsUser have a GUID for an ID?
* What is the ClaimsUser's GUID ID?

I think maybe the first two could be combined, but the third should definitely not be.

Thanks for pulling me up on this, we need to try and get this stuff fixed properly, and this is a golden opportunity.

There is nowhere in the system where in any normal circumstance we should every have a GUID UserId which is not a real user's GUID.  This only needs to be checked at the front-end auth, and then after that never again.

Let's banish the Guid.Empty and Guid? thing completely from all the code, and make proper functions which do what they say, and not a bunch of secondary things.  I'm going to put in issue to cover this explicitly.

PR #482, opened 2016-02-23
2nd attempt at fix for #475

chrismdev:
2nd attempt at fix for #475

### DIFF
--- a/Bonobo.Git.Server.Test/Bonobo.Git.Server.Test.csproj
+++ b/Bonobo.Git.Server.Test/Bonobo.Git.Server.Test.csproj
@@ -161,6 +161,7 @@
     <Compile Include="TestCategories.cs" />
     <Compile Include="UnitTests\PasswordServiceTest.cs" />
     <Compile Include="UnitTests\GitAuthorizeAttributeTest.cs" />
+    <Compile Include="UnitTests\UserExtensionsTests.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\Bonobo.Git.Server\Bonobo.Git.Server.csproj">
--- a/Bonobo.Git.Server.Test/UnitTests/UserExtensionsTests.cs
+++ b/Bonobo.Git.Server.Test/UnitTests/UserExtensionsTests.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+using Bonobo.Git.Server;
+
+namespace Bonobo.Git.Server.Test.UnitTests
+{
+    [TestClass]
+    public class HelperTests
+    {
+        const string domainslashusername = @"domain.alsodomain\username";
+        const string usernameatdomain = "username@domain.alsodomain";
+
+        [TestMethod]
+        public void GetDomainFromDomainSlashUsername()
+        {
+            Assert.AreEqual("domain.alsodomain", domainslashusername.GetDomain());
+        }
+
+        [TestMethod]
+        public void StripDomainFromDomainSlashUsername()
+        {
+            Assert.AreEqual("username", domainslashusername.StripDomain());
+        }
+
+        [TestMethod]
+        public void GetDomainFromUsernameAtDomain()
+        {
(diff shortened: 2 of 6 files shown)


RedX2501:
We should always store the username as provided because only the user knows the format it'll accept. I'll try to find out why user@domain got converted.

chrismdev:
The issue was (for me anyway) that when authenticating via a browser this format worked:  username@domain.local but when git connected it used DOMAIN\username which failed. We don't seem to handle both.

RedX2501:
I'm having a different result. If i pass `user@domain` i get `user@domain` to authenticate...

RedX2501:
It seem that @larshg introduced a bug in e48d6247 it changed the authentication provider code when using windows auth. @chrismdev try reverting the changes done to GitAuthorizeAttribute in e48d6247 and it all should work fine again. I'll make a PR to your repo for you to test.

RedX2501:
@chrismdev Please try out the PR I have created.

chrismdev:
Success. That was it, thanks @RedX2501

larshg:
All this validation is not supposed to happen inside the membership service? - atleast it also tries to strip the domain etc etc.

I can't see the point in having the ADMembershipService if we don't use it?

But to handle the various usernames that can be provided - we might need to debug/test/improve the ADMembershipService :)

RedX2501:
It is the old story between authenication and identification.
Theoretically AuthenticationProvider is only supposed to tell you if you have access.
While the membershipservice should identify you. Tell me who you are in the system and to which groups you belong.
We don't have an ADAuthenticationProvider (i don't know why) so that's why the ADMembershipProvider does it.

We might need to rethink the architecture at that point. The correct way IMHO would be to remove auth from both membershipprovider and force AD users to use WindowsAuth, not sure what that would do to the FederationUsers.

larshg:
Yes, that might be true.

But the EF/Cookie authentication also uses the membershipservice right below the same way? ie validating the username / password.

And if you strip the domain from the username in the ADMembership validate method - that would also work.

This gives us only a single place where we do all this domain handling/various usernames for AD. But the entire method of ValidateUser might have to be split up - but then its the case for both AD end the EF provider?

At least I can reproduce the same error, so it does need to be corrected somewhere :)

RedX2501:
So i guess this is ready to merge? If nobody complains for 24 hours i'll merge.

willdean:
I don't know enough about how the AD/WinAuth stuff works to comment on the substance but 
:+1: on adding some tests :-)

`if(AuthenticationProvider is WindowsAuthenticationProvider) {blah blah } ` is a code smell as bad as Satan's underpants, but hopefully all that will get sorted if we do the #483 stuff.

larshg:
Is there any reason why we cannot strip the domain from the username in the ValidateUser method instead? and use that - until it gets reworked?

Atleast the validateUser does exactly the same thing as you are just to about to reapply? (Except removing the Domain from the user and hence fail to find the user through the AD)

chrismdev:
I wasn't expecting my code to be accepted. The function would be better written like this:

Bonobo.Git.Server/Security/ADMembershipService.cs
      ```
        {
            string domain = GetDomainFromUsername(username);
            UserModel usr = null;
            if (!IsUserPrincipalName(username))
            {
                using (PrincipalContext principalContext = new PrincipalContext(ContextType.Domain, ActiveDirectorySettings.DefaultDomain))
                using (UserPrincipal user = UserPrincipal.FindByIdentity(principalContext, username))
                {
                    // assuming all users has a guid on AD                 
                    usr = ADBackend.Instance.Users.FirstOrDefault(n => n.Id == user.Guid.Value);
                }
            }
            else if (!string.IsNullOrEmpty(username) && usr == null)
            {
                usr = ADBackend.Instance.Users.Where(x => x.Username == username).FirstOrDefault();
            }
            return usr;
        }
```

RedX2501:
@chrismdev since i foresee a lot of changes going on in the near future for {EF,AD}MembershipServices i'd say we leave it alone for now as it is working. If you really feel that that change would improve things, you are welcome to create an extra PR for it. Maybe with some tests?

PR #485, opened 2016-02-24
Strip domain from username else it won't be validated sucessfully by AD.

larshg:
Bug in StripDomain

can you test this @chrismdev - to see if this also fix your error?

### DIFF
--- a/Bonobo.Git.Server/Extensions/UserExtensions.cs
+++ b/Bonobo.Git.Server/Extensions/UserExtensions.cs
@@ -89,7 +89,7 @@ namespace Bonobo.Git.Server
             int delimiterIndex = username.IndexOf('@');
             if (delimiterIndex > 0)
             {
-                username = username.Substring(0, delimiterIndex - 1);
+                username = username.Substring(0, delimiterIndex);
             }
             delimiterIndex = username.IndexOf('\\');
             if (delimiterIndex > 0)
--- a/Bonobo.Git.Server/Security/ADMembershipService.cs
+++ b/Bonobo.Git.Server/Security/ADMembershipService.cs
@@ -34,6 +34,8 @@ namespace Bonobo.Git.Server.Security
             try
             {
                 string domain = GetDomainFromUsername(username);
+                username = username.StripDomain();
+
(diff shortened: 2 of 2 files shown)


Bonobo.Git.Server/Security/ADMembershipService.cs
```diff
@@ -89,14 +91,13 @@ public IList<UserModel> GetAllUsers()
         public UserModel GetUserModel(string username)
         {
             string domain = GetDomainFromUsername(username);
-            if (!IsUserPrincipalName(username))
```
larshg: @willdean and @RedX2501 - do you guys have any idea why we would skip getting the user, if its a UserPrincipalName? I guess thats in a form of username@domain.domainend - Is it because that was the primary key at some point :P ?

Bonobo.Git.Server/Extensions/UserExtensions.cs
```diff
@@ -89,7 +89,7 @@ public static string StripDomain(this string username)
             int delimiterIndex = username.IndexOf('@');
             if (delimiterIndex > 0)
             {
-                username = username.Substring(0, delimiterIndex - 1);
```
RedX2501: I have found this bug and fixed it in https://github.com/chrismdev/Bonobo-Git-Server/commit/4d40b539a358fda9ddae2d663477a9b842d9dc09. It's awaiting to be merged.

Bonobo.Git.Server/Security/ADMembershipService.cs
```diff
@@ -34,6 +34,8 @@ public ValidationResult ValidateUser(string username, string password)
             try
             {
                 string domain = GetDomainFromUsername(username);
```
RedX2501: Maybe we should remove this helper function and use username.GetDomain().

Bonobo.Git.Server/Extensions/UserExtensions.cs
larshg: I know - but this should be an alternative solution. And hence I also needed to change this :)

larshg:
Not necessary now the other has been merged...

PR #487, opened 2016-02-27
Add various features

RedX2501:
1. I have added a pseudolocalization to afrikaans so we can see if we are missing something.
2. Implemented displaying a copy link button to the repository listing.
3. Implemted dropdown to quickly jump from repo to repo
4. Allow < in descriptions and names where reasonable.
5. Add renaming feature


RedX2501:
Since im using raw with the rest i felt safer doing the extra encode. 

I don't think we are opening any xss stuff because i tested it and always saw the fields properly escaped into the proper html escape chars. But I'm not so experienced with all the methods of doing xss Injection.

RedX2501:
Yeah, i am going to add the tests tomorrow. 

From recollection alone i see that it is backwards. Sorry. 

I looked into that and figured doing twice the lookup was too much. Also i think that in 99% of the cases we want a case insensitive match of the repo name. 

With this set of commits i was looking more into a feedback regarding the layout of the dropdown menu and the copy link buttons. 

But I'm glad for the feedback regarding the code and will add the tests and fix the things you mentioned.

willdean:
Copy link buttons - I'm no UX/Graphic designer, but I think they might look better aligned in some way - so either on the left of the repo name or over to the right by the the edit/delete buttons.  I think we need to differentiate between the two types of buttons too, though I don't know how.  JOOI, what is the use-case for the personal urls?

I like the repo pull down - is it in the best place on the page, being between the repo name and the repo-specific buttons?    I don't know where a better place is though - top left?  top right?   Is this something which could actually be on the title bar so it's available from everywhere?

RedX2501:
All Points you raised are valid and have crossed my mind. 

1. Personal url is nice when cloning for the first time as it saves you from typing your username.
2. Adding the dropdown to every view means every view must populate some ViewBag or similar. I could not find a practical way of doing this automatically. Maybe you have more luck then i have had when googling.

I'll try positioning the buttons on their own columns and see how they look.

willdean:
For (2) the most popular answer here might be relevant http://stackoverflow.com/questions/5453327 - (this is NOT the accepted answer).  I haven't tried it and there is some stuff in the comments about it affecting ViewBag in action methods which would be a pain if it was true.

willdean:
What about combining icon-user with the copy icon to make the personal login button?

RedX2501:
I have improved the location of the icons. I haven't been able to figure out how to get all buttos look the same but I like the way they look now. Buttons one color and links the other.

I tried setting the dropdown on the upper banner but gave up when trying to move PopulateRepoGoToList out of the Controlle. Too many dependencies in there. I think it would be a nice location for it. But it needs too much refactoring.

willdean:
Icons look much better positioned.  I don't know about background colour, etc, not really my field!

I had a quick go at pulling PopulateRepoGoToList out of the controller - if you do RepoRepo.GetAllRepositories rather than GetIndexModel it comes out much more easily - don't know if that's helpful.

```
   List<SelectListItem> PopulateRepoGoToList(Guid id, RequestContext context)
        {
            var firstList = RepositoryRepository.GetAllRepositories()
                    .OrderBy(x => x.Name.ToLowerInvariant())
                    .GroupBy(x => x.Group ?? Resources.Repository_No_Group);
            List<SelectListItem> items = new List<SelectListItem>();
            var u = new UrlHelper(context);
            var groups = new Dictionary<string, SelectListGroup>();
            foreach (var grouped in firstList)
            {
                SelectListGroup group = null;
                string key = grouped.Key;
                if (!groups.TryGetValue(key, out group))
                {
                    group = new SelectListGroup();
                    group.Name = key;
                    groups[key] = group;
                }
                foreach (var item in grouped)
                {
                    var slt = new SelectListItem
                    {
                        Text = item.Name,
                        Value = u.Action("Detail", "Repository", new { id = item.Id }),
                        Group = group,
                        /* This does not seem to work.
                         * If someone can figure out why we can remove the "Go to repository"
                         * from the drop down creation. */
                        Selected = item.Id == id,
                    };
                    items.Add(slt);
                }
            }
            return items;
        }


```

RedX2501:
Yeah but now you are skewing all repos. Not just the ones the user has access to.

willdean:
Oh yes, good point.  Most of the guts of GetIndexModel ought to be somewhere like the IRepositoryPermissionService implementations anyway.  I'll look at getting that under test and then moving it.  If that was more easily available outside RepositoryController, then that would help.

RedX2501:
Updated dropdown location.

@willdean I noticed that you can only query one Permisison at a time. This means the index will no longer show repositories where you can only administer but not pull (if that exists.)

willdean:
@RedX2501 Although I was joking about people putting the permission levels in a special order about doing evil comparisons against them, I think there is an implicit order, and each level implies everything below it.

* Everyone who is an admin has permission to push or pull (after all, they can always go and set that permission anyway), 
 * Everyone who can push must also be able to pull.

So effectively Admin > Push > Pull.    If you're an admin on a repo, CheckNamedUserPermission always grants pull & push permission.

I think your drop-down should use the Pull permission.  (it may do, I haven't looked at the code yet)

RedX2501:
@willdean Any idea what could have caused the errors? Does anything get executed in parellel with ForAllGits?

willdean:
Nothing happens in parallel in ForAllGits - unfortunately I can't see the failures now because I think I can only see the most recent pass.

RedX2501:
@willdean you can go to history and see the one that failed. Make sure you are logged in.

Issue #368, opened 2015-11-18
Federation Auth: Unauthorized


stupiduglyfool:
Bonobo Version: 5.0.1.0.
Issue: Attempting to login with a federation setup takes me to an unauthorized page.

Details:

I already have an adfs environment setup and working, it is currently operational with other relying parties and works well with for example IdentityServer3.

I have created a relying party for bonobo, I presume it requires a passive setup, the federation guide does not specify what the Passive URL should be or which adfs profile to use.. these are the values I have used:

ADFS 2.0 Profile
No token encryption certificate
Enable support for WS-Fed Passive Protocol URL: https://git.fqdn/bonobo.git.server 
RelyingPartyIdentifier: urn:bonobo

I have these values in my web.config:

<add key="AuthenticationProvider" value="Federation" />       	  
<add key="MembershipService" value="ActiveDirectory" />
<add key="ActiveDirectoryDefaultDomain" value="mydomain.co.uk" />
<add key="ActiveDirectoryBackendPath" value="~\App_Data\ADBackend" />
<add key="ActiveDirectoryMemberGroupName" value="Developers" />
<add key="ActiveDirectoryTeamMapping" value="Team1=Developers" />
<add key="ActiveDirectoryRoleMapping" value="Administrator=Senior Developers" />	
	    
<add key="FederationMetadataAddress" value="https://adfs.fqdn/federationmetadata/2007-06/federationmetadata.xml" />
<add key="FederationRealm" value="urn:bonobo" />

When I goto https://git.fqdn/bonobo.git.server I am prompted by adfs for my credentials.. which I enter credentials of a user I know to be in the above groups, but unfortunately I get taken to an Unauthorized page. Additionally there doesn't seem to be any useful information in any logs.. but I admit I cannot see any info in the documentation about increasing log levels.

Any help or guidance would be appreciated.

stupiduglyfool:
So I've grabbed the latest code and got digging. 

During debugging it seems I am successfully logged in, I have an Identity, IsAuthenticated is true.. but I do not have any roles despite having the config above set up as per the documentation.

WebAuthorizationAttribute.OnAuthorization IsInRole(Definitions.Roles.Member)

Always returns false, additionally ADRoleProvider is never called, therefore there is nothing to relate my identity to the Member or Administrator roles and thus I get forwarded to unauthorized. Looking through the commit history of other Role providers I see this has had a bit of a makeover recently, so this maybe an oversight.

[...]

PR #488, opened 2016-02-27
Change claims names to be more standard

willdean:
This is a proposal to change the claims names we use, to make them slightly more standard, and allow more straightforward integration with ADFS, (and hopefully other external login systems should we ever offer them.)

This PR uses the following claims:

* **Name** = Username (unique in the membership database)
* **Upn** (UserPrincipleName) = Alternative location for username - we don't set this claim, but if Name isn't set, we'll look here for a username instead.  
* **NameIdentifier** - Guid ID for user.  We can accept this being either a string Guid, or being a base64 encoding of a Guid, which is what the ADFS ObjectGuid is.
* **GivenName**, **Surname**, **Email** - As you'd expect
* **Role** - Unaffected

This is based on work done for #368 with @stupiduglyfool, though it's not exactly what he's been testing.  Prior to these changes, he had to have a rather complex an un-intuitive mapping of ADFS claims to Bonobo claims, and a scripted Guid translation.  With this change, it should be much simpler.

Within Bonobo, it doesn't matter what claims we use, because we're packing stuff in and out of the claims identity ourselves anyway - this is just to help with other services which provide claims information externally, and to avoid unpleasant surprises like #471.

Thoughts...?

PS I don't like the 'DisplayName' hack which I put in - we could add a DisplayName claim, and thus pass it right through from AD/ADFS rather than constructing it ourselves, which is bound to risk internationalisation problems, etc.   It's actually only used for the heading text at the top of the page.

### DIFF
--- a/Bonobo.Git.Server.Test/UnitTests/UserExtensionsTests.cs
+++ b/Bonobo.Git.Server.Test/UnitTests/UserExtensionsTests.cs
@@ -1,7 +1,6 @@
 ﻿using System;
+using System.Security.Claims;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
-
-using Bonobo.Git.Server;
 
 namespace Bonobo.Git.Server.Test.UnitTests
 {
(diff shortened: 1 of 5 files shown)


RedX2501:
You talk about getting the username from upn but you never set it, how is that supposed to work? Does adfs set it?

willdean:
As far as I know, a ClaimsIdentity gets built somewhere in the WsFederation OWIN stuff from the ticket provided by ADFS.

If you look at https://github.com/jakubgarfield/Bonobo-Git-Server/issues/368#issuecomment-187757256 you can see @stupiduglyfool setting up the claims which ADFS issues.

RedX2501:
LGTM

PR #492, opened 2016-03-01
RFC Unified repo permissions

### LINKED ISSUE #491: Repository Permissions
At present, `IRepositoryRepository` has a method called `GetPermittedRepositories`, which is used by `RepositoryController` to get a list of repositories that the specified user is allowed to view - this is used to populate the index view (and @RedX2501's new pull-down list).

This method was semi-duplicated between AD and EF, but I combined them last week, so there's currently only one version of `GetPermittedRepositories`, used for both AD and EF.

Separately, there is a method on `IRepositoryPermissionService` called `HasPermission`, which takes a user Id and a Repo Id, and tells you if that user has permission on that specific repo.

Intuitively, you might think that `GetPermittedRepositories` would give the same answer as `GetAllRepositories().Where(...HasPermission(..))`.

However, this is not so - they differ in subtle ways:  

1. `GetPermittedRepositories` does not check to see if the specified user is the system admin - so this has to be done separately in `RepositoryController`.  However, it _will_ return all repos marked for anon access.

2. HasPermission **does** check to see if the specified user is system admin, but **doesn't** check if the repo is marked for anon access.  So when HasPermission is used, it tends to be coupled with a separate test for AllowsAnonymous

In addition to these rules being implemented multiple times in slightly different ways, the simple boolean concept of 'I have permission on this repo' is inadequate.   The rules for push and pull might be different for anonymous users, for example.  And permission to 'administer' repositories has to be checked separately.   This code ends up spread across multiple other places in the code, and the corner cases and interactions are very hard to check.   I came into the project because of bugs in the permissions logic, and there have been loads of issues around this sort of stuff, as well as feature requests which are hard to implement with confidence.  

I propose something like the following:

1. We define some levels of access request for a repository - I suggest 'Administer', 'Push' and 'Pull', but there might be something else we need.
2. We agree that Guid.Empty is the user ID of the anonymous user (I think it effectively is anyway)
3. We implement a single HasPermission method in the RepositoryService, which has a signature like this:
`bool HasPermission(Guid userId, Guid repositoryId, RepositoryAccessLevel level)`

Whenever you want to know if a user can access a repository, you call this routine, with the type of access you require.  (Mostly it would be 'Pull').   There's no function which returns an RepositoryAccessLevel in any way - it's just a level you ask for if you need.

Every kind of check for repo admin, system admin, team membership, anonymous rules, etc, _everything_ happens though this one call.   `GetPermittedRepositories` would then just be a helper which filtered the full list of repos through this call.

There are various requests like  #434 and #311 and even #77 which would be much easier if one could be more confident about this stuff, and issues like #390 would be much less likely to happen if the code wasn't so difficult work work on.

The nice think about pulling this code out of places like GitController, RepoController and the various auth attributes would be that it would be much easier to test for a consistent set of rules.

Thoughts?

tl;dr - I'd like to consolidate the currently scattered-about repo permission logic into a single point.

willdean:
Fixes #491 - I am pleased with how GitController and WebAuthorizeRepositoryAttribute have come out - this is what I was hoping for.

Feedback welcomed


willdean:
OK, I had a bit of an issue with the AnonPush test:

In the master branch, AnonPush triggers a bug in the GitAuth attribute which is handled badly and so leads to the usual `http://localhost:20000/Integration.git/info/refs not valid: is this a git repository?` which you get when we serve up nonsense to the git client.

In this PR branch, I've fixed a number of problems to do with auth on anon repos, which now means that when you push to an anon repo, Git asks you for login credentials - if you can log in you can push, if you can't login, then you won't be able to push.

Unfortunately, when running under the test harness, when the Git client asks for credentials, we get this error `bash: /dev/tty: No such device or address.....(on and on and on, three lines of stuff)..`.  So I've changed the match-string in the resources to match this.   Hope this is OK.

BTW, I think if we ever see `/info/refs not valid: is this a git repository?` when we've spoken to the proper Git URL then we should say that's a bug - I don't think there's any situation where we'd expect that now.

RedX2501:
Altogether I really like the interface proposed and the implementation.

One thing I noticed is that when we do an anonymous `git push` the `User.Id` in SecureGetInfoRefs will try to do a conversion of a `null` Guid leading to the branch with the trace instruction in UserExtensions.cs:47. Maybe on anonymous interactions we should set the claim for guid (NameIdentifier) to empty on GitAuthorizeAttribute?

I'm still not sure I like being asked about a password when the anonymous push is not possible. Maybe returning 401 and leaving the user to provide his username on the URL might be a clearer option of what happened. But it is more trouble changing the URL after you have cloned... Now that you know my opinion on this I will accept whatever option you choose.

willdean:
> One thing I noticed is that when we do an anonymous git push the User.Id in SecureGetInfoRefs will try to do a conversion of a null Guid leading to the branch with the trace instruction in UserExtensions.cs:47. Maybe on anonymous interactions we should set the claim for guid (NameIdentifier) to empty on GitAuthorizeAttribute?

This was really just me not thinking about anon users at all when I put that trace in there - I've taken it out now. If you have something in the claim which can't be parsed, it will still log, but if it's empty we assume you're the anonymous user.

>I'm still not sure I like being asked about a password when the anonymous push is not possible. Maybe returning 401 and leaving the user to provide his username on the URL might be a clearer option of what happened. But it is more trouble changing the URL after you have cloned... Now that you know my opinion on this I will accept whatever option you choose.

I never ever use the personal URLs myself, so I never consider the opportunities.  But I would say that I think 5.1.1 works like my branch does - if you push at an anon-repo on the normal URL, you get asked for credentials.  I'm also not sure that the current master would work even with a personal URL, unless the URL contained the password too, because the 'AllowsAnonymous' test causes all processing of the auth header to be skipped.

RedX2501:
LGTM

PR #502, opened 2016-03-04
Repo local push and better info on unknown repo Id.

RedX2501:


RedX2501:
There is no localization for the enum yet. I'll see if i can figure it out tomorrow.

willdean:
Re these failures in a the directory deletion in the integration tests - Have a look at https://github.com/willdean/Bonobo-Git-Server/commit/693006a74597802ebb8cfcb362a28da330fbcb2d which will be part of a new PR soon - this has a beefed-up Delete routine which I have previously found is needed to cope with transient file locking by other things.

RedX2501:
Does this LGTM to anyone?

willdean:
Yup LGTM.

Issue #500, opened 2016-03-03
Rework Contributors/Administrators Permissions Dialog


hmmwhatsthisdo:
Currently, the Contributors/Administrators permissions dialog is two checkbox-lists with identical contents.

It would likely be better if these two lists were zipped together, with the checkboxes on the right-hand side a la the Windows file permissions dialog.

Moreover, the current list is not sorted in any meaningful fashion. Sorting names alphabetically by first name or by last name with "Last, First" display format (Again, like the Windows file permissions dialog) would likely be much more usable.

willdean:
I'd say that the combining the lists might not make it for the next release, but the sorting should be easily fixed.

hmmwhatsthisdo:
What about doing "Last, First"-style display/sorting?

willdean:
The current master displays 'Display Name', which is constructed from "GivenName Surname" - we can definitely sort that on Surname without any problem.

hmmwhatsthisdo:
Alright. Would it also be possible to fall back to the AD displayName and/or userPrincipalName attributes if the given/surname are both blank? (One of our admin accounts only had a displayName attribute set in AD and showed up on Bonobo with a blank name)

willdean:
The difficulty with taking DisplayName from AD at the moment would be where to store it within Bonobo, because we don't have a DisplayName field in the internal structures.  (This could be fixed at some point)

Falling back to Username (which would be the UserPrincipalName on an AD user) for DisplayName is a good idea, because showing a blank is completely hopeless.

hmmwhatsthisdo:
Using the userPrincipalName attribute would work fine. (Perhaps, once storing the displayName attribute attribute is possible, it would be worth it to show the uPN on hovering over a username so people with multiple accounts can distinguish between them easily)

RedX2501:
Yeah i definitely wanted to show the username when hovering but i found out no easy way of doing that with mvccheckboxlistfor do i left it for later

I thought i had sorted it by first name but maybe it got lost in refactoring. 

I'm no big fan of last, first. That is an American thing. We should make it configurable.

willdean:
I am wondering if we do need to address the sorting issue here, because we probably used to sort 'by accident' on username, which was reasonable, and now we sort 'by accident' on GUID, which is not really very helpful.

I am included to make GetAllUsers sort on the SortName property I added in #503 - any thoughts on this?

chilversc:
[...]

PR #503, opened 2016-03-04
Fall back to Username if DisplayName cannot be formed

willdean:
This is an uncontroversial, I think, change to DisplayName to make sure that if people don't have a Given/Surname set then they get the username shown rather than a blank.  Fixes one complaint in comments on #500, which I think is important for the next release, as blank Display Names are a bit of a disaster.

This also adds a 'SortName' which isn't used anywhere at present but could be used to sort the users by  (it's not intended for display)


willdean:
I'm going to merge this because I think it's safe, and if we don't like the sort-order stuff it doesn't matter because it's not used at the moment

PR #504, opened 2016-03-04
[RFC] Push to create

### LINKED ISSUE #376: Scripted Repository Creation

willdean:
Fixes #376.  Was started in PR #460 

@RedX2501 I made quite a big change to the Integration tests to provide a GitResult class rather than using the tuple (all the Item1,Item2,Item3 stuff confuses me).   I hope this isn't offensive to you - if it causes you hassle with changes you're making in parallel, I'm happy to do any tedious merge work which arises.

I feel happier in tests I write to see the asserted conditions more explicitly in the test method rather implied in the helper methods, so I've tried to allow the result to be returned to the test method instead of being consumed in the helper, but then added code in the result object to make testing stuff easy.   I would value your opinion.

I cannot get the two tests for push-to-create to work here on my machine - using the admin:admin@ style of URL with the server instance which SpecsForMVC spins-up always suffers an auth error (the client never sends the HTTP basic auth header).    I don't know why - I can use the same kind URL with the same code if I run it in Visual Studio, and the admin/admin account is OK for web-login into the SpecForMVC server instance.   However, they do work OK on Appveyer.

This PR also has a more robust DirectoryDelete routine in the integration tests, which may help with intermittent test failures.

### DIFF
--- a/Bonobo.Git.Server.Test/IntegrationTests/GitResult.cs
+++ b/Bonobo.Git.Server.Test/IntegrationTests/GitResult.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+﻿using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Bonobo.Git.Server.Test.IntegrationTests
 {
@@ -29,21 +30,9 @@ namespace Bonobo.Git.Server.Test.IntegrationTests
 
         public void ErrorMustMatch(MsysgitResources.Definition resource, params object[] args)
         {
-            string matchString;
-            if (args.Length > 0)
-            {
-                matchString = string.Format(Resources[resource], args);
-            }
-            else
-            {
-                matchString = Resources[resource];
-            }
-            var expected = matchString.Trim();
+            var expected = string.Format(Resources[resource], args).Trim();
             var actual = StdErr.Trim();
-            if (expected != actual)
-            {
-                Assert.Fail("Git operation StdErr mismatch - expected '{0}', was '{1}'", expected, actual);
-            }
+            Assert.AreEqual(expected, actual, "Git operation StdErr mismatch");
         }
     }
 }
--- a/Bonobo.Git.Server.Test/IntegrationTests/GitResult.cs
+++ b/Bonobo.Git.Server.Test/IntegrationTests/GitResult.cs
@@ -1,0 +1,49 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Bonobo.Git.Server.Test.IntegrationTests
+{
+    public class GitResult
+    {
+        public string StdErr { get; set; }
+        public string StdOut { get; set; }
+        public int ExitCode { get; set; }
+        public MsysgitResources Resources { get; set; }
+
+        public bool Succeeded
+        {
+            get { return ExitCode == 0; }
+        }
+
+        public bool AccessDenied
+        {
+            get { return !Succeeded && StdErr.Contains(Resources[MsysgitResources.Definition.AuthenticationFailedError]); }
+        }
+
+        public void ExpectSuccess()
+        {
+            if (!Succeeded)
+            {
+                Assert.Fail("Git operation failed with exit code {0}, stderr {1}", ExitCode, StdErr);
+            }
+        }
+
+        public void ErrorMustMatch(MsysgitResources.Definition resource, params object[] args)
+        {
+            string matchString;
+            if (args.Length > 0)
+            {
+                matchString = string.Format(Resources[resource], args);
+            }
+            else
+            {
+                matchString = Resources[resource];
+            }
+            var expected = matchString.Trim();
+            var actual = StdErr.Trim();
+            if (expected != actual)
+            {
+                Assert.Fail("Git operation StdErr mismatch - expected '{0}', was '{1}'", expected, actual);
+            }
+        }
+    }
+}
--- a/Bonobo.Git.Server.Test/IntegrationTests/MsysgitIntegrationTests.cs
+++ b/Bonobo.Git.Server.Test/IntegrationTests/MsysgitIntegrationTests.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Diagnostics;
 using System.Collections.Generic;
 using System.Linq;
+using System.Linq.Expressions;
 using SpecsFor.Mvc;
 using OpenQA.Selenium;
 
@@ -11,7 +12,7 @@ using Bonobo.Git.Server.Controllers;
 using Bonobo.Git.Server.Models;
 using System.Text;
 using System.Threading;
-
+using Bonobo.Git.Server.Test.IntegrationTests;
 using Bonobo.Git.Server.Test.IntegrationTests.Helpers;
 using OpenQA.Selenium.Support.UI;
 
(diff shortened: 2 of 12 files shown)


Bonobo.Git.Server.Test/IntegrationTests/GitResult.cs
```diff
@@ -0,0 +1,49 @@
...
+                Assert.Fail("Git operation failed with exit code {0}, stderr {1}", ExitCode, StdErr);
+            }
+        }
+
+        public void ErrorMustMatch(MsysgitResources.Definition resource, params object[] args)
+        {
+            string matchString;
+            if (args.Length > 0)
+            {
+                matchString = string.Format(Resources[resource], args);
```
RedX2501: Will you get an error if args.length == 0? Else i'd remove the branch as it is a pointless optimization IMHO.

Bonobo.Git.Server.Test/IntegrationTests/GitResult.cs
```diff
@@ -0,0 +1,49 @@
...
+            }
+            else
+            {
+                matchString = Resources[resource];
+            }
+            var expected = matchString.Trim();
+            var actual = StdErr.Trim();
+            if (expected != actual)
+            {
+                Assert.Fail("Git operation StdErr mismatch - expected '{0}', was '{1}'", expected, actual);
```
RedX2501: The format for all other Assert comparisons is actual, expected. Maybe we should also do it here for consistency. Why not use Assert.AreEqual here anyway?

Bonobo.Git.Server.Test/IntegrationTests/MsysgitIntegrationTests.cs
```diff
@@ -243,8 +246,45 @@ public void NamedPushToAnonRepo()
...
+                CreateAndAddFiles(git);
+
+                RunGitOnRepo(git, "push origin master").ErrorMustMatch(MsysgitResources.Definition.RepositoryNotFoundError, RepositoryUrlWithCredentials);
+            });
+        }
+
+        [TestMethod, TestCategory(TestCategories.ClAndWebIntegrationTest)]
+        public void PushToCreateIsAllowedIfOptionIsSet()
+        {
+            ForAllGits((git, resource) =>
```
RedX2501: I thought you wanted to use your new Git object here instead of the tuple? Did i miss something?

Bonobo.Git.Server.Test/IntegrationTests/MsysgitIntegrationTests.cs
```diff
@@ -378,15 +423,31 @@ private void InitAndPullRepository(string git, MsysgitResources resource)
...
+
+        private void InitRepository(string git)
+        {
+            RunGitOnRepo(git, "init").ExpectSuccess();
+            RunGitOnRepo(git, String.Format("remote add origin {0}", RepositoryUrlWithCredentials)).ExpectSuccess();
+        }
+		
+        private void InitAndPushRepository(string git, MsysgitResources resource)
+        {
+            RunGitOnRepo(git, "init");
```
RedX2501: You should alse check for success in all these calls to early detect any problems.

Bonobo.Git.Server.Test/IntegrationTests/MsysgitIntegrationTests.cs
```diff
@@ -378,15 +423,31 @@ private void InitAndPullRepository(string git, MsysgitResources resource)
...
             var result = RunGitOnRepo(git, "pull origin master");
 
-            Assert.AreEqual(String.Format(resource[MsysgitResources.Definition.PullRepositoryError], RepositoryUrlWithoutCredentials), result.Item2);
+            Assert.AreEqual(String.Format(resource[MsysgitResources.Definition.PullRepositoryError], RepositoryUrlWithoutCredentials), result.StdErr);
+        }
+
+
+        private void InitRepository(string git)
+        {
+            RunGitOnRepo(git, "init").ExpectSuccess();
```
RedX2501: Are you sure that only checking the exit code is enough?

Bonobo.Git.Server/App_GlobalResources/Resources.resx
```diff
@@ -682,6 +682,12 @@
   <data name="Settings_Global_AllowUserRepositoryCreation_Hint" xml:space="preserve">
     <value>If checkbox is unchecked, only administrators can create repositories.</value>
   </data>
+  <data name="Settings_Global_AllowPushToCreate" xml:space="preserve">
```
RedX2501: I think that as all other repo configuration stuff, there should be a repo local setting for it also. Or even only a repo local setting. If you wish i can create the update scripts for it.

Bonobo.Git.Server/Controllers/GitController.cs
```diff
@@ -80,6 +102,36 @@ public ActionResult SecureReceivePack(String repositoryName)
             }
         }
 
+        private bool TryCreateOnPush(string repositoryName)
+        {
+            DirectoryInfo directory = GetDirectoryInfo(repositoryName);
```
RedX2501: Don't you have to combine this with the current setting for repositorypath?

Bonobo.Git.Server.Test/IntegrationTests/GitResult.cs
willdean: It's not really an opptimisation, if you don't have any args then we don't try to expand the format string, sonyou can inclue curly brackets without a problem.  But you're probably right it's unnecessary.

Bonobo.Git.Server.Test/IntegrationTests/MsysgitIntegrationTests.cs
willdean: I didn't do that mod, I wasn't sure what you thought about it

Bonobo.Git.Server.Test/IntegrationTests/MsysgitIntegrationTests.cs
willdean: Yes

Bonobo.Git.Server.Test/IntegrationTests/MsysgitIntegrationTests.cs
willdean: I think it's a reliable check for success, when it's not the main assertion of the test method.

Bonobo.Git.Server/App_GlobalResources/Resources.resx
willdean: If you can write a per-repo setting for permission to create the repo which contains the permission option, then be my guest :-).   Funny enough i went down the same line of thought before realising the problem.

Bonobo.Git.Server/Controllers/GitController.cs
willdean: Yup this is wrong, thanks

willdean:
Thanks for the feedback, will have a goes at this tomorrow

RedX2501:
I noticed that we have not being updated the changelog file properly. Do you mind adding an entry for this PR?

Bonobo.Git.Server.Test/IntegrationTests/MsysgitIntegrationTests.cs
RedX2501: Oh sorry. I'm totally fine with any change that will increase separation of concerns, increases code readability and reduces code duplication. We should only try to keep helper methods in the IntegrationTestHelpers (maybe they only exist in my branch yet) so that we can reuse them in the not Git tests.

Maybe we should make an own PR for it.

Bonobo.Git.Server/Controllers/GitController.cs
willdean: Oh no, it's correct actually - GetDirectoryInfo does this.

PR #505, opened 2016-03-04
Allow repo to be saved by server-admin without a repo admin being set

### LINKED ISSUE #501: Git Admins shouldn't need to mark their own accounts as repo administrators when editing permissions
Right now, when a server-wide admin opens the Edit dialog for a repository, they have to tag their account as an administrator within the administrator list for that repo in order to be able to save changes. Not doing so throws the error "You can't remove yourself as an administrator".

This seems a bit counter-intuitive, as it shouldn't matter whether or not a server admin is an admin over a specific repository - they're going to have access regardless.

willdean:
Fixes #501 

I tried to write some integration tests, but got stuck at trying to get SpecsForMvc to find all the admin checkboxes and turn them off.

### DIFF
--- a/Bonobo.Git.Server.Test/IntegrationTests/Controller/RepositoryControllerTests.cs
+++ b/Bonobo.Git.Server.Test/IntegrationTests/Controller/RepositoryControllerTests.cs
@@ -139,6 +139,21 @@ namespace Bonobo.Git.Server.Test.IntegrationTests.Controller
 
         }
 
+/*
+        [TestMethod]
+        public void RepositoryCanBeSavedBySysAdminWithoutHavingAnyRepoAdmins()
+        {
+            var repoId = ITH.CreateRepositoryOnWebInterface(app, "Repo");
+
+            app.NavigateTo<RepositoryController>(c => c.Edit(repoId));
+            // TODO - need to find all the admin checkboxes here and turn them off
+            app.FindFormFor<RepositoryDetailModel>()
+                .Field(f => f.PostedSelectedAdministrators).SetValueTo("0")
+                .Submit();
+
+        }
+*/
+
     }
 }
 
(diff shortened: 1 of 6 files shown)


RedX2501:
I created a PR on your repo with checkboxselection example.

RedX2501:
I noticed that we have not being updated the changelog file properly. Do you mind adding an entry for this PR?

willdean:
@RedX2501 This branch is currently failing test on AppVeyor in RepoAdminCannotRemoveThemselves at the point where it tries to create a user. This works when I run the test on its own locally.

I suspect this might be because the user already exists as a hang-over from an earlier test - does each test method in the xxxControllerTests get a clean start, or could they be interacting? I notice lots of tests are doing 'DeleteRepository' at the end, which they shouldn't need to do if the framework is cleaning up.

RedX2501:
As far as I understand the framework it will copy the contents of the bin folder only once when starting. I also think that it starts the server only once. This means that whatever is left in the DB after each test is still there.

To circumvent problems with left behind repois i have made the CreateRepos function now return an IDisposable. This should probably be done for the CreateUser function too.

willdean:
Better post-test cleanup is good, but it might also not be unreasonable to have some kind of local-only 'reset database' call into Bonobo, which could be used by the test framework to initialise the database into a known state before each test.

We could then perhaps have both an 'empty' (admin only, no repos) condition, and a 'test mode' condition, which pre-loaded a configuration with a repos and a couple of non-admin users, and perhaps a team.   If we had a standardised test setup, then it would save having to write that stuff into individual tests, and it would be much faster to run each test if the framework didn't need to painfully step through endless web-pages setting it up.

Do you think we could come up with a standardised setup which would be a helpful starting point for writing tests?

PR #507, opened 2016-03-05
[RTM] Add environment variables as requested. Fix #495

### LINKED ISSUE #495: Do we have any option to read user group name or user roles in git hooks?
We would like to restrict users to delete remote branch based on user role/ users belong  to a group.

Can we read that information from git server side hooks?

RedX2501:
I have no idea on how to test this.


RedX2501:
Please merge if noone has any findings.

PR #508, opened 2016-03-05
[RTM] Linkify per repo

RedX2501:


RedX2501:
Please merge if there are no more points open.

PR #509, opened 2016-03-05
Enforce sqlite connection string must contain binary guid = false.

### LINKED ISSUE #506: BinaryGUID=False needs to be detected/enforced.
It's well established that for the guid update you need to have BinaryGUID=False; in the sqlite connection string, otherwise bad things happen.

I am thinking that we need to detect/enforce this before/during the AddGuidColumn update when people start upgrading to 6.0.0, because otherwise we could end up in a world of misery if people hang on to their existing web.configs.

Either that or fix the code to work with binary guids in the sqlite code - I'm not sure what's involved in that.

RedX2501:
Fix #506


RedX2501:
You get a YSOD because the error handler has problems with non http exceptions.

RedX2501:
I think we can merge this?

willdean:
Yes, go for it.

PR #512, opened 2016-03-06
[RTM] Verify as you type

RedX2501:
While entering username, team name or repository name while editing, it will check if they are unique.

### DIFF
--- a/Bonobo.Git.Server.Test/Bonobo.Git.Server.Test.csproj
+++ b/Bonobo.Git.Server.Test/Bonobo.Git.Server.Test.csproj
@@ -144,11 +144,13 @@
     <Compile Include="MembershipTests\ADTests\ADPermissionServiceTest.cs" />
     <Compile Include="MembershipTests\ADTests\ADRepositoryRepositoryServiceTest.cs" />
     <Compile Include="MembershipTests\ADTests\ADTestSupport.cs" />
+    <Compile Include="MembershipTests\ADTests\ADTeamRepositoryTests.cs" />
     <Compile Include="MembershipTests\EFTests\EFPermissionServiceTest.cs" />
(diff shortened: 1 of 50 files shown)


willdean:
I believe I have sorted out the routing / ID problem - I've committed this to my fork, commit willdean:59bb52644

However, it still fails test because of some kind of test interaction (the tests run individually but not all together.)

I'd like to try it on my database-resetting branch, but now that it's rebased onto the idisposable stuff, it's a really difficult merge.

I'm not sure what the best thing to do is now - I've tried making a new branch and cherry-picking the good bits of this one, but there are things like team-tests in a2447ad which depend on the IDisposable stuff which don't come across very well.

Not sure the best way to proceed here - I think this is a brillant feature, really nice, but this doesn't seem like a good PR now it has 17 commits in it, most of which are not really anything to do with the feature.

I'll have another go at just cherry-picking the bare minimum, and see if I can at least get a test pass, then we can work out what to do.

RedX2501:
Please merge if there are no more points open.

willdean:
Why is #5e47740 in this PR?

RedX2501:
Because that is the test for when you type a wrong regex without submitting the form.

willdean:
I'm confused between the linkification and the 'verify as you type' PRs here - I thought 'verify as you type' was about renaming users/teams/repos.   It seems like we've got all the 'per repo linkification' stuff on this PR too

RedX2501:
Hmm i guess github doesn't handle pr which depend on pr very well. This was rebased on top of the other pr.

willdean:
It's just showing the difference between the branch and master, so if you've rebased on onto the other, it makes sense that one includes both.

Maybe just merge the other one and then this one will make more sense

PR #514, opened 2016-03-08
[RTM] Make all tests dispose Repos, Teams and Users automatically

RedX2501:
This creates Disposable for Repositories, Teams and Users, which should delete themselves after each test.

### DIFF
--- a/Bonobo.Git.Server.Test/IntegrationTests/Controller/RepositoryControllerTests.cs
+++ b/Bonobo.Git.Server.Test/IntegrationTests/Controller/RepositoryControllerTests.cs
@@ -68,93 +68,90 @@ namespace Bonobo.Git.Server.Test.IntegrationTests.Controller
         public void CreateDuplicateRepoNameDifferentCaseNotAllowed()
         {
             var reponame = "A_Nice_Repo";
-            var id1 = ITH.CreateRepositoryOnWebInterface(app, reponame);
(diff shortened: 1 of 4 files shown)


RedX2501:
Please merge this as fast as you can. I need this to make the other two PRs finally work properly.

willdean:
GH reckons this will merge OK - is there a problem with doing that?

willdean:
I'm not going to make a big deal of this, but I actually don't think this basic premise of this PR is a good idea.  I think the solution to the test-interaction problem is good setup, not more complex clean-up after tests run.

My problem with this approach is:
* It makes the tests more complex (you need to remember the using, and there's more indent, etc)
* It makes the tests run more slowly, because the cleanup takes time
* A mistake in one test (forgetting a using) may cause a failure not in that test but in a different one.

But it's better than manual deletion of stuff, which is what we had before...

RedX2501:
I agree that this is not fast. The using is actually totally optional. At the end of the test stuff should be destroyed.

Doing the merge should be fine. I rebased it making sure that it shoud be a clean fast forward.

Please merge whenever you feel like it.

RedX2501:
Im not even sure why I'm using an IDisposable interface. I could make them delete at the destructor.

I saw that you already have a branch with the other option. I guess I'll just wait for that.

PR #516, opened 2016-03-08
Possible source of ideas - verify as you type 

willdean:
@RedX2501 I'm not really proposing you merge this as-is, because it makes a complete pain of your other branches which branch off disposables, but this does contain what I did to get the id stuff working in the validation controller (in 9f25ca5), along with the database reset stuff - I couldn't get the tests to pass in the other branch.

I'm 99% that some of the recent problems with the tests are because the client-side uniqueness-validation is creating a delay which causes various problems to operations which occur after the 'submit()' in the tests - I think that's possibly a bug in SpecsForMvc, because I think Submit is supposed to block until the page has been redrawn.  I had to mess around with my code which watches for validation information - adding retries and delays, and I know you added a delay too in an earlier commit.


RedX2501:
I think all things you did here sholud no be in the proper branch. I'm closing this for now.

PR #517, opened 2016-03-09
[RTM] Reset framework

RedX2501:
This is a proposal for the reset framework for integration tests.

I added a transform to the publish process that should be used when testing. It creates an extra key AllowDBReset in the web.config. Without that it is also not possible to reset the database using the DatabaseResetCode

### DIFF
--- a/Bonobo.Git.Server.Test/Bonobo.Git.Server.Test.csproj
+++ b/Bonobo.Git.Server.Test/Bonobo.Git.Server.Test.csproj
@@ -143,6 +143,7 @@
     </Otherwise>
   </Choose>
   <ItemGroup>
+    <Compile Include="IntegrationTests\IntegrationTestBase.cs" />
     <Compile Include="IntegrationTests\SharedLayoutTests.cs" />
     <Compile Include="IntegrationTests\GitResult.cs" />
     <Compile Include="MembershipTests\ADTests\ADBackendStoreTest.cs" />
--- a/Bonobo.Git.Server.Test/Bonobo.Git.Server.Test.csproj
+++ b/Bonobo.Git.Server.Test/Bonobo.Git.Server.Test.csproj
@@ -37,20 +37,11 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release-Api|AnyCPU' ">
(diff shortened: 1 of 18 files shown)


willdean:
:+1: I really like what you've added to this

This test failure has been appearing intermittently ever since I wrote the test:

> Failed   PushToCreateIsAllowedIfOptionIsSet 
Error Message:  Assert.Fail failed. Git operation failed with exit code 128, stderr fatal: repository 'http://admin:admin@localhost:20000/Integration.git/' not found

I think what might be happening is a race between the 'Submit' call finishing (and so the global anon-push option gets set), and the Git call being made.

I thought this couldn't happen, because the docs on Submit() suggest that it blocks waiting for the page to reload, but I have a feeling that's buggy.  (Ref all those problems we've had in the last 24hrs with the unique-name validation making the tests break)

I've just added `AssertThatNoValidationErrorOccurred` checks after the various Submit() calls in MsysGitIntegrationTests - I'm trying it on appveyor now on my fork.

We might consider having a `ITH.SubmitExpectingSuccess()` method which does both the Submit and the `AssertThatNoValidationErrorOccurred`, and then get in the habit of using that in every situation where we're doing a 'normal' submit.

The test failure where mstest just crashes with an error and a suggestion to look at a dmp file may just be mstest stupidity - we could perhaps consider nUnit or something else if we can't avoid the mstest problems..

RedX2501:
I'm still not happy here because something odd is happening.

The transfomration should only occurr on the Test build configuration so all tests should have failed immediately especially since we configure AppVeyor to run the Release build.

Can someone help me figure this out? This  is the first time I'm having to do these transformations on my own.

willdean:
I don't think we need to have a different build do we?   It's SpecsForMVC which is doing the web.config transformation.  You just need to make the transformation file available to SpecsFor.  Appveyor shouldn't care

RedX2501:
You are right. I was under the impression that i tried that once locally and it had failed.

willdean:
@RedX2501 Can you cherry-pick https://github.com/willdean/Bonobo-Git-Server/commit/3a7d80bd5313fbe274585ab0f8a85b1f45f1635f into this branch - that should improve the reliability further.

This has passed on my appveyor, but I think you've rebased the branch and I don't want to tread on what you're doing.

RedX2501:
Done. Please merge whenever you feel like it.

willdean:
If we get a test pass I'll merge it as soon as

PR #520, opened 2016-03-12
[RTM] Added view for tags. Fix #304

### LINKED ISSUE #304: Releases / tag page wanted
It would be nice if bonobo-git-server can come with a releases / tag page like github. This allow us to easily view all the releases (tag) on a page with information for each tag.

RedX2501:
I noticed the commits page does not look very good when MANY tags are present.

### DIFF
--- a/Bonobo.Git.Server.Test/IntegrationTests/GitResult.cs
+++ b/Bonobo.Git.Server.Test/IntegrationTests/GitResult.cs
@@ -20,19 +20,21 @@ namespace Bonobo.Git.Server.Test.IntegrationTests
             get { return !Succeeded && StdErr.Contains(Resources[MsysgitResources.Definition.AuthenticationFailedError]); }
         }
 
-        public void ExpectSuccess()
+        public GitResult ExpectSuccess()
         {
             if (!Succeeded)
             {
                 Assert.Fail("Git operation failed with exit code {0}, stderr {1}", ExitCode, StdErr);
             }
+            return this;
         }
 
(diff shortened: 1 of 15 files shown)


willdean:
I can't test this today, but it seems very low risk in terms of the rest of the project, so I reckon you should merge it if you think it works OK.

RedX2501:
@willdean I needed to touch the routing to be able to test the generated links. Optional arguments should not be used in Controllers as it makes finding the generated Urls impossible. They should go as default ids in the routing table.

RedX2501:
@willdean it seems that there is no upn in the claims which is now needed by the AntiForgeryToken. Not sure how it was working on your branch...

willdean:
@RedX2501 In my 5.1.2 branch there's this line in global.asax.cs:

`            AntiForgeryConfig.UniqueClaimTypeIdentifier = ClaimTypes.Upn;`

Because 5.1.x versions used ClaimTypes.Upn for the username.

In 6.x this will need to be 

`            AntiForgeryConfig.UniqueClaimTypeIdentifier = ClaimTypes.NameIdentifier;`

(Actually, I think that's the default, we we probably don't need the line at all - I haven't tried it)

willdean:
I've just pushed a change to remove this line from master

willdean:
OK, I'm screwing this AF stuff up on master, sorry - I'll get it sorted out an merge it back properly.  Leave it with me.

willdean:
OK, AF stuff sorted and merged now, all tests passing

PR #523, opened 2016-03-15
Fixed bundle names causing 404 in some server configurations

mrmashal:
As stated [here](http://stackoverflow.com/a/16216752/1785552), I encountered the same issue for Bonobo Git Server, on my production server, causing the pages rendered without style because of 404 responses for bundle requests. This PR fixes this issue.
Thank you


RedX2501:
Lgtm, @willdean what do you think?

willdean:
I just had a quick look at a large MVC app I maintain here, and we have bundles called things like 'AllScript' and 'AllCss'.  I can't recall if they're like that to avoid IIS confusion with .css and .js paths, but it's clear that it works OK with non-dotted names.

I'm :+1 to merge this, I can't see a downside.

Issue #521, opened 2016-03-15
Cleanup the routing table


RedX2501:
For #520 I needed to add an entry to the routing table. I noticed that we have many out of date entries in there.

We definitely need to clean the table up.

The question is are there any advantages in doing the routing using MVC attributes since we are using version 5.0?

willdean:
I like attribute-routing for stuff which is used in a small number of places (like the GitController for example).  I've only ever used them mixed with a 'traditional' routing table, so I don't really know whether it would be good to replace everything with them, but I'm :+1: on introducing them for new stuff.

A related useful (perhaps!) task would be to review where URLs are being hand-assembled rather than built with the route-aware helpers and replace that where possible with code which understands the routing.  (I know the Git URLs are done like this, and it isn't exactly a work of art...)

KY4U:
Hello,

I believe my issue may be related to the routing table.  My installation is on an Azure webapp.  After pulling and publishing the latest code from 03/14/16 I'm getting 401 errors for urls that reference /info/refs?service=git-upload-pack

The problem is that the generated url contains only the webapp name and not the full url.  Here is an example:  https://mywebappname:80/git/my_repo.git/info/refs?service=git-upload-pack

If using the Azure domain then the url would be: https://mywebappname.azurewebsites.net:80/git/my_repo.git/info/refs?service=git-upload-pack

or if the actual Azure custom domain were used:
https://mycustomdomain.com/git/my_repo.git/info/refs?service=git-upload-pack

Please advise.
Thanks

willdean:
Is this URL with the ?service= part on it coming out of a Git error message?   

Can you paste both the error message and the Git URL which is on the 'Repository Details' page within Bonobo?

KY4U:
The incorrect url I'm getting from the error log directory in Azure.

HTTP Error 401.0 - Unauthorized
You do not have permission to view this directory or page.
Requested URL:  https://nccos-coastalscienceadmin-web:80/git/Powershell_MergeXML.git/info/refs?service=git-upload-pack

Repository URL: https://admin.coastalscience.noaa.gov/git/Azure-Powershell.git

Basically when I attempt to get to /Repository/Detail/asldkjalsdkfj I end up getting a Bonobo error page and I find this error in the log files.

Thanks for the help.

willdean:
@KY4U Sorry, I am really struggling to get a grip on what's happening here.

Can you confirm how much of this is true:

[...]

PR #524, opened 2016-03-15
Allow DisplayEnum to format invalid enum values without crashing

willdean:
This prevents a crash on the Repository Details page if the database contains invalid values for AllowAnonymousPush - although this problem is just down to unreleased changes pre v6.0.0, it's still good if a general purpose enum formatter performs no worse than `.ToString()`

Addresses one of the problems raised in #521 (nothing to do with routing)


RedX2501:
The Patch looks good. 

What I'm thinking is if we should force on save a valid value instead of keeping the invalid one.

PR #526, opened 2016-03-16
Verifies that the passed enum is a defined value 

RedX2501:
Complementary to #524 this will only allow valid values to be posted when creating or editing repos.

### DIFF
--- a/Bonobo.Git.Server.Test/IntegrationTests/Controller/RepositoryControllerTests.cs
+++ b/Bonobo.Git.Server.Test/IntegrationTests/Controller/RepositoryControllerTests.cs
@@ -269,5 +269,41 @@ namespace Bonobo.Git.Server.Test.IntegrationTests.Controller
             var input = app.Browser.FindElementByCssSelector("input#LinksRegex");
             Assert.IsTrue(input.GetAttribute("class").Contains("input-validation-error"));
         }
+
+        [TestMethod, TestCategory(TestCategories.WebIntegrationTest)]
+        public void AnonymousPushModeNotAcceptInvalidValueWhenEditingRepo()
+        {
+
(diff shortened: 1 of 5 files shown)


willdean:
I don't think this is worth doing.  It's an incredibly unusual case - there's no normal way of a user injecting this wrong value into the database via a web-page.  It adds a lot of clutter to the controller action to very little benefit, IMO.

RedX2501:
He can change it by editing the select field value. That's how I do the testing.

willdean:
Do you mean changing it by playing with the developer tools to change the POSTed field values?

RedX2501:
No, i change the form in selenium with setAttribute and then send the form.

willdean:
OK, I get that for testing, but is that the kind of thing we generally protect against (people injecting illegal form values into the browser using extra tools)?   I'd have thought there would be loads of stuff which could be done using those sort of techniques which would break various form posts.

I don't have a strong feeling about this really, and I'd sound a bit deranged trying to construct an argument against server-side form validation.

Is there a way to do this with attributes on the viewmodel property which then feeds directly into the ModelState without having special code in the action method?   So then it would be like the other validation, and would automatically apply whenever that viewmodel was used?

RedX2501:
I can create an attribute for that.

Generally you try to prevent any kind of non authorized change. That's why you have the validation attributes on server side. In case the user disables javascript for instance on their side, or modify the forms structure before sending, or plainly forges a POST with all kinds of mischievous data.

willdean:
Yes, I get the arguments for server-side validation in general, it was just that protecting against repo-admins who want to forge form data didn't seem a high priority!   But the general principle is good.

This might be relevant http://stackoverflow.com/questions/26425242

willdean:
LGTM

PR #527, opened 2016-03-16
Correct the initial value for AnonymousPush column. Fix #525

### LINKED ISSUE #525: AddRepoPushColumn sets column to invalid value
I've just posted #524 to deal with a crash reported in #521 which was caused by an invalid enum value in AllowAnonymousPush.  I had thought this had come from some intermediate data in the database left over from partial updates, but I've now realised that it's coming from AddRepoPushColumn which sets the new column to '3', which is not a valid member of the associated enum.

I think this should be a valid member, not an invalid one, and I suspect 'No' (0) is the correct default. 

Any thoughts, @RedX2501 ?

RedX2501:


willdean:
Looks OK

PR #528, opened 2016-03-16
Remove duplicate code

padremortius:
Remove duplicate code from  Bonobo.Git.Server/Git/GitService/GitServiceResultParser.cs


willdean:
Can you expand on what this does?   Doesn't it completely change the processing of the stream to remove a read?

I think there is much to improve in some of that Git code, but it would be really good to get it under test first.

PR #530, opened 2016-03-17
[RFC] Fix update guid

RedX2501:
I tried creating tests for the update paths but it seems that all InitialCreateScript versions were to same up until this release.

Yet the upgrade fails when going from 1.2.0 to the latest version. I must be missing something.

I used following script to get all the release InitialCreateScript versions that were released:

```
#!/usr/bin/env bash

set -e
set -u

while IFS= read -r info; do
	sha=${info%% *}
	tag=${info##*/}
	echo "sha ${sha:0:6} tag $tag"
	while IFS= read -r script; do
		if ! [[ $script =~ "SqlServer" ]]; then
			echo "i $script"
			git checkout -f $sha
			filename=$(basename $script)
			ext=${filename##*.}
			mv $script "./${filename%.*}_${tag}_${sha:0:6}.$ext"
		fi
	done < <(git ls-tree -r --name-only $sha | grep -i InitialCreateScript.cs)
done < <(git show-ref --tags)
```

### DIFF
--- a/Bonobo.Git.Server.Test/Bonobo.Git.Server.Test.csproj
+++ b/Bonobo.Git.Server.Test/Bonobo.Git.Server.Test.csproj
@@ -171,6 +171,7 @@
     <Compile Include="TestCategories.cs" />
     <Compile Include="UnitTests\PasswordServiceTest.cs" />
     <Compile Include="UnitTests\GitAuthorizeAttributeTest.cs" />
+    <Compile Include="UnitTests\DatabaseUpdateTests.cs" />
     <Compile Include="UnitTests\UserExtensionsTests.cs" />
     <Compile Include="UnitTests\UserModelTest.cs" />
   </ItemGroup>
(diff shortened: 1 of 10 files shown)


jakubgarfield:
Awesome, that fixes the issue. I couldn't spot the trailing comma :)

RedX2501:
@willdean  Can you have a look why this is failing? It's in the push to create test. Completely unrelated to my changes. Probably some odd timing thing again...

willdean:
I'll take a look - this has been failing intermittently for ever.  I thought I might have fixed it by waiting for completion on the options post, but it appears that isn't enough.

RedX2501:
@willdean @jakubgarfield I reordered the update scripts with this patch to keep update working from older versions.

Now it will only insert the inital data after all updates have been done.

Please have a look and commit if you think it is ok.

willdean:
@RedX2501  - The add-administrator stuff was broken for SqlServer.

Can you think of any reason why the code to add an administrator has to be hand-coded Sql?  Could we not just add it using the membership service, which could then be common to Sqlite/sqlserver and would be far cleaner code?

RedX2501:
I guess that before our Introduction of DoAction all update scripts had were strings. 

I think we could change it now. I think that's stuff for a new PR for 6.1?

RedX2501:
@willdean Can we merge this?

willdean:
@RedX2501  - I think so.

PR #538, opened 2016-03-25
AD Controller event log spamming

ThaStealth:
Fixed issue where the AD validation would cause a lot of events in the AD Controller event log everytime a login happend and SSL is not used.

### DIFF
--- a/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
+++ b/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
@@ -118,7 +118,7 @@ namespace Bonobo.Git.Server
                     var adUser = UserPrincipal.FindByIdentity(pc, username);
                     if (adUser != null)
                     {
-                        if (pc.ValidateCredentials(username, password))
+                        if (pc.ValidateCredentials(username, password, ContextOptions.Negotiate))
                         {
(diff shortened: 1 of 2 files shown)


RedX2501:
What does the added option do exactly? 

Can you please add a space between the comma and the option?

ThaStealth:
Space added, 
Without the option the function will first try to use an SSL connection (which fails because the certificate is not in the AD cert store). After this fails (and generates a logging event) the function will attempt a regular connection.With it will directly try a regular connection

RedX2501:
You probably also need to change Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs:121:

ThaStealth:
Good catch, changed it!

RedX2501:
@willdean This patch will fix the problems because now the PrincipalService is allowed to choose the connection method. Altough in my opinion the correct fix is to add the SSL certificate to the machines certificate storage. Should we merge this?

@ThaStealth Why don't you add the certificate to your windows certificate storage. Using non-ssl connections for authentication should not be done.

willdean:
@RedX2501  I don't know enough about AD to know whether this is a good or a bad thing.  I don't expect the authentication is insecure even without SSL though, so I don't think that is a concern.

PR #539, opened 2016-03-26
Made all 3 auth providers have same sign in function. Set the cookie …

### LINKED ISSUE #537: remember me doesn't
When signing in I select "Remember Me". If I then close or leave the browser idle then when I return to Bonobo Git Server I'm no longer signed in. Looking at the cookies for the site their expiriy is set to session.

I'm using the cookies authentication provider and the internal membership provider.

RedX2501:
…to persistent. Increased timeout to 3 days from 1 hour. Fix #537

See http://stackoverflow.com/questions/20226016/cookieauthenticationoptions-expiretimespan-does-not-work#comment42772288_20259788 for details.


willdean:
Does this always set a persistent login cookie, even if people *don't* say "Remember Me?"

RedX2501:
That is a good question. I have not modified the code from the login action in the controller, so whatever did the remember me magic before should still work. I'll try it out.

RedX2501:
@willdean It seems that the remember me function never worked but i guess nobody cared because if you use windows auth or fed auth or AD membership you did get logged in automatically.

Please have a look and merge if you think it is ok.

PR #564, opened 2016-05-10
AD: Support groups containing users from multiple domains in forest

mr-miles:
Hi,

These changes are working here against our AD setup.  Changes fall into two camps:
- for user operations, pass around domain+username, and consistently use the given domain or the default to look up the user;
- for UpdateUsers in ADBackend, once you have looked up the group look up the user against the expected domain

I don't have good visibility on how to make it more generic within the structure of the app.  Also, should the groups not be in the default domain then AD will still have problems.  The only solution to that would be to require all the config groups to be in domain\group format which has backwards compat issues that you might not want.

Two other things:
- I cannot get the personal url to work.  I am not sure whether that is because of AD or the newer version of the code but I don't have a readily-available setup that I'm confident of.  I suspect double @ in the url is causing the problem.
- When I upgraded my instance, I got a error from sqllite about adding a guid column when it went to update.  I just removed the db and it worked, but looking through the code I think it was because CopyUsers was not specifically using UserPrincipalName and AD was getting confused.  However I had no old data to test

Hope that all makes sense, let me know if not and I can adjust things

### DIFF
--- a/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
+++ b/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
@@ -110,17 +110,17 @@ namespace Bonobo.Git.Server
         private bool IsWindowsUserAuthorized(HttpContextBase httpContext, string username, string password)
         {
             var domain = username.GetDomain();
-            username = username.StripDomain();
+            var justUsername = username.StripDomain();
             try
             {
                 using (PrincipalContext pc = new PrincipalContext(ContextType.Domain, domain))
                 {
-                    var adUser = UserPrincipal.FindByIdentity(pc, username);
+                    var adUser = UserPrincipal.FindByIdentity(pc, justUsername);
                     if (adUser != null)
                     {
-                        if (pc.ValidateCredentials(username, password, ContextOptions.Negotiate))
+                        if (pc.ValidateCredentials(justUsername, password, ContextOptions.Negotiate))
                         {
-                            httpContext.User = new ClaimsPrincipal(new ClaimsIdentity(AuthenticationProvider.GetClaimsForUser(username.Replace("\\", "!"))));
+                            httpContext.User = new ClaimsPrincipal(new ClaimsIdentity(AuthenticationProvider.GetClaimsForUser(username)));
                             return true;
                         }
                     }
--- a/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
+++ b/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
@@ -110,6 +110,7 @@ namespace Bonobo.Git.Server
         private bool IsWindowsUserAuthorized(HttpContextBase httpContext, string username, string password)
         {
             var domain = username.GetDomain();
+            if (String.IsNullOrWhiteSpace(domain)) domain = ActiveDirectorySettings.DefaultDomain;
             var justUsername = username.StripDomain();
             try
             {
--- a/Bonobo.Git.Server/Security/ADMembershipService.cs
+++ b/Bonobo.Git.Server/Security/ADMembershipService.cs
@@ -93,21 +93,15 @@ namespace Bonobo.Git.Server.Security
 
         public UserModel GetUserModel(string username)
         {
-            if (!UsernameContainsDomain(username))
-            {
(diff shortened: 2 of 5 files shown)


Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
```diff
@@ -110,17 +110,17 @@ private bool IsUserAuthorized(string authHeader, HttpContextBase httpContext)
...
                 using (PrincipalContext pc = new PrincipalContext(ContextType.Domain, domain))
                 {
-                    var adUser = UserPrincipal.FindByIdentity(pc, username);
+                    var adUser = UserPrincipal.FindByIdentity(pc, justUsername);
                     if (adUser != null)
                     {
-                        if (pc.ValidateCredentials(username, password, ContextOptions.Negotiate))
+                        if (pc.ValidateCredentials(justUsername, password, ContextOptions.Negotiate))
                         {
-                            httpContext.User = new ClaimsPrincipal(new ClaimsIdentity(AuthenticationProvider.GetClaimsForUser(username.Replace("\\", "!"))));
```
RedX2501: This was definitely a bug. Still left around from a patch I introduced. Thanks for spotting this.

Bonobo.Git.Server/Security/ADMembershipService.cs
```diff
@@ -93,21 +93,15 @@ public IList<UserModel> GetAllUsers()
...
-            {
-                using (PrincipalContext principalContext = new PrincipalContext(ContextType.Domain, ActiveDirectorySettings.DefaultDomain))
-                using (UserPrincipal user = UserPrincipal.FindByIdentity(principalContext, username))
-                {
-                    // assuming all users have a guid on AD
-                    return ADBackend.Instance.Users.FirstOrDefault(n => n.Id == user.Guid.Value);
-                }
-            }
-            else if (!string.IsNullOrEmpty(username))
+            var domain = username.GetDomain();
```
RedX2501: Does this work if no domain has been provided in the username?

RedX2501:
@willdean I had a look at it. Nothing looks too suspicious on it. I have no way of testing it. How about you?

RedX2501:
@mr-miles Why do you have double @ in your names? The stripDomain and getDomain functions will certainly have problems with that.

Can you post an example?

Bonobo.Git.Server/Security/ADMembershipService.cs
mr-miles: Yes - next line uses the default domain if no explicit domain specified

mr-miles:
sure - that is just a hunch btw.  my personal url looks like this:

http://M.Waller%40company.com@uk-pro-qcoreapp/Bonobo.Git.Server/JapanDashboard.git

I was suspicious that the encoding of the @ or the fact there are two was preventing it from working but not a huge amount of evidence other than that everything else is working ok.

larshg:
I'm currently trying to test this and I have problem that I have used a local url (drs.lan) for our AD server, but using our public domain for our email/principle name? ie. user@ourcompany.dk.

So when updating the users, the domain is looked up to be ourcompany.dk, but the server does not respond to this.

But this might be more a failure of running a server properly than bonobo to handle this? Or if the found domain is different than the default domain - it should try both?

Hope I'm making sense @mr-miles

mr-miles:
I'm not an AD expert, so I don't have a whole lot to go on as to how
flexible is "normal".  I thought our "multi domain" structure was
complicated, but I think running the server on a different name from the
one it manages is probably quite a lot more custom!

Could you set up AD in your hosts file?  Or in a set of self-contained VMs?

Failing that, I think I saw from MSDN that you can pass in the dns name in
one of the constructors, or give it some sort of base query, but that
probably is a customisation the setup doesn't expose at the moment.

FWIW, since the most recent commit, I've had no feedback from our users
that its not working or seen errors in the logs.

Miles

RedX2501:
@mr-miles The double @ can be troublesome if it does not get encoded to the GetDomain function  So maybe you could set a breakpoint there and see what you get from it. Do you log in with `user%40email.com@company.com` from the web? Does that work?

larshg:
@mr-miles can you test my #644 to see if it works in your environment?

willdean:
I'm closing this as I don't think it's going to get merged now without a lot of work (and it may already be supported by more recent AD work).

Issue #570, opened 2016-05-23
Problem with markdown rendering


lkho:
It seems the current markdown library MarkdownDeep is not rendering some of the text correctly, especially code blocks. Would it be better if changed to the CommonMark.NET implementation? Since I think it conforms to the standard and compatible with popular sites such as GitHub.

npbenjohnson:
It would be nice to get this, or atleast an option of markdown providers since the markdown generators I use generally make something more functional and nicer looking with github markdown enabled.

RedX2501:
Can you maybe post some screenshots to evaluate if it is worth the trouble?

lkho:
sample md: https://github.com/mikecao/flight/blob/master/README.md
v5.1.1:
![5 1 1](https://cloud.githubusercontent.com/assets/7780243/16226552/86f02308-37de-11e6-9c1e-8e35e8ed6f88.png)

v6.0.0 up to lkho@607429cb7a56f5f09186fc714233a2c0a73483a2:
![6 0 0](https://cloud.githubusercontent.com/assets/7780243/16226557/8a24d1cc-37de-11e6-90ca-2df878c80d74.png)

RedX2501:
It seems you already have implemented it in your branch, can you create a pull request with the changes? Please also update the changelog file.

PR #579, opened 2016-06-26
replace MarkdownDeep with CommonMark.NET

lkho:
#570


RedX2501:
@jakubgarfield @willdean I'd commit this if none of you has anything against it.

willdean:
I think I've taken a sacred vow not to hold any opinion on Markdown, ever.  Go for it.

jakubgarfield:
👍

PR #580, opened 2016-06-27
[WIP] Update the ADBackend information when updating from pre 6.0

RedX2501:
Hopefully this fixes the problems for ActiveDirectory users that lost all their data when updating to the unreleased 6.0 version.

To test if it works: Checkout/install version 5.2 create some repos/teams. Checkout this branch and see if you can see the repos/teams with all their informations/members correctly.

@willdean Can you give this a spin? 
@chrismdev Can you give this a spin?

### DIFF
--- a/Bonobo.Git.Server/Data/Update/ADBackend/UpdateADBackend.cs
+++ b/Bonobo.Git.Server/Data/Update/ADBackend/UpdateADBackend.cs
@@ -141,6 +141,7 @@ namespace Bonobo.Git.Server.Data.Update.ADBackendUpdate
             {
                 var repo = repoitem.Value;
                 var newrepo = new Models.RepositoryModel();
+                newrepo.Id = Guid.NewGuid();
                 newrepo.Name = repo.Name;
                 newrepo.Group = repo.Group;
                 newrepo.Description = repo.Description;
@@ -246,7 +247,7 @@ namespace Bonobo.Git.Server.Data.Update.ADBackendUpdate
                 }
 
                 var domainuser = UserPrincipal.FindByIdentity(dc, u.Username);
-                if (domainuser != null)
+                if (domainuser != null && domainuser.Guid.HasValue)
                 {
                     u.Id = domainuser.Guid.Value;
                 }
--- a/Bonobo.Git.Server/Data/Update/ADBackend/UpdateADBackend.cs
+++ b/Bonobo.Git.Server/Data/Update/ADBackend/UpdateADBackend.cs
@@ -1,0 +1,156 @@
+﻿using Bonobo.Git.Server.Data.Update.Pre600ADBackend;
+using System.IO;
+using Bonobo.Git.Server.Helpers;
+using System.Collections.Generic;
+using System.DirectoryServices.AccountManagement;
+using Bonobo.Git.Server.Configuration;
+using System;
+
+namespace Bonobo.Git.Server.Data.Update.ADBackendUpdate
+{
+    public class Pre600UpdateTo600
+    {
+        // Before 6.0.0 the mapping was done via the name property. After that the Guid is used.
+        public static void UpdateADBackend()
+        {
+            // Make a copy of the current backendfolder if it exists, so we can use the modern models for saving
+            // it all to the correct location directly
+            var dir = PathEncoder.GetRootPath(ActiveDirectorySettings.BackendPath);
+            var bkpdir = PathEncoder.GetRootPath(ActiveDirectorySettings.BackendPath + "_pre6.0.0");
+            if (Directory.Exists(dir))
+            {
+                if (!Directory.Exists(bkpdir))
+                {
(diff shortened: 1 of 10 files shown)


RedX2501:
@willdean Did you find any time to test this?

willdean:
I haven't had a chance to set up the AD stuff again - I'm flat-out on a deadline for the end of July, hopefully after that I'll be a bit freer

willdean:
I have tried to test this here and cannot currently get my 5.2 installation to talk to my AD server, so I can't do the 'before' bit of the upgrade.  I am hoping that @Trapulo can repro this in #602, but I will keep trying here anyway.

Bonobo.Git.Server/Data/Update/ADBackend/UpdateADBackend.cs
```diff
@@ -19,9 +21,104 @@ public static void UpdateADBackend()
...
+
                 if (!Directory.Exists(bkpdir))
                 {
-                    new Microsoft.VisualBasic.Devices.Computer().FileSystem.CopyDirectory(dir, bkpdir);
+                    {
+                        new Microsoft.VisualBasic.Devices.Computer().FileSystem.CopyDirectory(dir, bkpdir);
+                    }
+                    // Make sure the Computer() instance has been disposed se we can delete the source directory
+                    // If anyone knows a better way of doing this, please use it.
+                    GC.Collect();
```
willdean: Can we use Directory.Move here instead of this VB stuff?

Bonobo.Git.Server/Data/Update/ADBackend/UpdateADBackend.cs
```diff
@@ -0,0 +1,156 @@
...
+            foreach(var roleitem in roles)
+            {
+                var role = roleitem.Value;
+                var newrole = new Models.RoleModel();
+                newrole.Name = role.Name;
+
+                newrole.Id = Guid.NewGuid();
+                var members = new List<Guid>();
+                foreach (var member in role.Members) {
+                    members.Add(users[member].Id);
```
willdean: The update fails here on my machine because I somehow have a role Member who isn't actually a real user, so the users[member] call fails, causing the whole update to abort.  We should probably use users.TryGetValue instead to be safe, and perhaps we should use more low-level exception handlers (with logging, of course) to make sure that as much of the update as possible completes, even if there are data-integrity issues along the way.

Bonobo.Git.Server/Data/Update/ADBackend/UpdateADBackend.cs
```diff
@@ -19,9 +21,104 @@ public static void UpdateADBackend()
             var bkpdir = PathEncoder.GetRootPath(ActiveDirectorySettings.BackendPath + "_pre6.0.0");
             if (Directory.Exists(dir))
             {
+                var mustConvert = false;
```
willdean: Instead of enumerating all the directories, then string matching their names, what about having a methods like 'IsDirectoryLoadable<ObjectType>(string backendDirectory, string subDirectoryName)' and calling it four times?

```
bool IsDirectoryLoadable<T>(string backendDirectory, string subDirectoryName)
{
  var x = JsonConvert.DeserializeObject<T>File.ReadAllText(Path.Combine(backendDirectory, subDirectoryName));
  return x.Id != Guid.Empty;
}
```

Something like that (I haven't tried it)

willdean:
I have had a go at this - I wrote various review comments and questions - I think I do now have a setup where I can easily try the update process.

Bonobo.Git.Server/Data/Update/ADBackend/UpdateADBackend.cs
RedX2501: I don't remember my logic for choosing copy over move, I guess there was none as I delete the directly afterwards. So I guess we can go for moving.

RedX2501: But I would not got for Guid. I'd go for the DateTime.Now as it makes it easier to find the last backup of a tried update.

Bonobo.Git.Server/Data/Update/ADBackend/UpdateADBackend.cs
RedX2501: You are right.

Bonobo.Git.Server/Data/Update/ADBackend/UpdateADBackend.cs
RedX2501: Yeah, no idea why I didn't use a function directly.

willdean:
@RedX2501 Thanks for reviewing my review!  If you're OK with it, I'll make some of the suggested changes, then I think we should merge this, because even when it fails it's no worse than the current situation.  Hopefully we have someone in #602 who is also able to test this.

RedX2501:
Sounds perfect for me. I'm currently very occupied by my work to do any changes on a short term. Thanks for being able to do the changes.

PR #584, opened 2016-07-08
Fixes Windows Authentication in GitAuthorizeAttribute

los93sol:
This change fixes Windows authentication so that when it is enabled the user is not prompted for credentials, instead they are passed via Basic authentication.

In addition, this change modifies the user check to pass their credentials with the domain the same way the check happens from the web interface.

### DIFF
--- a/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
+++ b/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
@@ -9,6 +9,8 @@ using Bonobo.Git.Server.Data;
 using Bonobo.Git.Server.Security;
 
 using Microsoft.Practices.Unity;
+using Microsoft.Owin.Security;
+using Bonobo.Git.Server.Owin.Windows;
 
 namespace Bonobo.Git.Server
 {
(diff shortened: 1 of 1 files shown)


Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
```diff
@@ -67,11 +69,25 @@ public override void OnAuthorization(System.Web.Mvc.AuthorizationContext filterC
...
-                    filterContext.Result = new HttpStatusCodeResult(HttpStatusCode.Unauthorized);
-                    return;
+                    if (AuthenticationProvider is WindowsAuthenticationProvider)
+                    {
+                        AuthenticationProperties authenticationProperties = new AuthenticationProperties()
+                        {
+                            RedirectUri = httpContext.Request.Path
+                        };
+
+                        httpContext.Request.GetOwinContext().Authentication.Challenge(authenticationProperties, WindowsAuthenticationDefaults.AuthenticationType);
```
RedX2501: This means that after Basic Auth their git client will try their username and password from windows? If that fails it'll ask the user for username and password?

larshg:
I have been testing this lately and I can't get it work with NTLM - also according to [this] post git doesn't support NTLM - but only Basic auth.
However you can set up proxy things inbetween. Or you can use the git credential manager to save passwords to avoid logging in more than once.

larshg:
@los93sol I'm closing this as it doesn't seem to correct anything that would actually work? If you think otherwise please reopen and supply additional information.

larshg:
I just need to test this a bit more after I have setup Git credential manager to handle windows authentication - then this might be necessary... At least some of it!

larshg:
I tried it again, but it keeps returning 401 and enters UnauthorizedResult in gitcontroller ?

larshg:
Closing due to inactivity

PR #600, opened 2016-09-14
Fixed issue commits pagination

ozielguimaraes:
The pagination always show the first 10 commits.


larshg:
It seems like this PR is broken? At least in only shows that a font is changed and I doubt that has something to do with pagination - @ozielguimaraes Can you find the original changes?

latop2604:
I pushed the same fixes previously. This is why the diff show nothing relevant.
The woff2 line in csproj is already there : https://github.com/jakubgarfield/Bonobo-Git-Server/blob/master/Bonobo.Git.Server/Bonobo.Git.Server.csproj#L533

Issue #605, opened 2016-10-14
should Bonobo have a Captcha on the login page


68681395:
in my cases , i offen deploy bonobo to public to internet, so , it will received a lot of attacks. so I think captcha is helpfull to some body like me.

68681395:
i am add the smart captcha function in my fork, and if you dont like this function , you can disable this captcha function in Global Setting page.

PR #606, opened 2016-10-20
Add a smart captcha when logon

68681395:
#605 completed the smart captcha funtion

if you didn't need it, you can diable it on Global Seting Page.


willdean:
This doesn't build at the moment because of this error:

MvcCaptcha\DefaultSmartCaptcha.cs(15,49): error CS0103: The name 'nameof' does not exist in the current context 

(i.e. we're not using VS2015 in the build, so you can't use nameof)

holiiveira:
Congratulations to those involved, this functionality is interesting. Is there any possibility of this feature being available also in removing a repository?

68681395:
@holiiveira There is absolutely possibility be available in removing a repository.  you only add the captcha input to View(x.cshtml) and add the attribute(ValidateMvcCaptcha) to your action of the controller. Then it can work well.

68681395:
I had fixed code, remove the 'nameof' to adapt IDE below vs2015. please help to review.

68681395:
Unfortunately, I only completed both English and Chinese resource file, and others resources need more contributes to complete it.

larshg:
@68681395 Isn't the majority of the code from https://tsharp.codeplex.com/? - for me it seems like its a entire implementation of a captcha adding to the source of bonobo - which I think is inappropriate to maintain in this project. It should be possible to use a package instead, like http://www.nuget.org/packages/RecaptchaNet/ or maybe googles implementation http://recaptchamvc.apphb.com/ ?

@willdean and @RedX2501  what do you think?

willdean:
I'm not really in favour of any of this - it seems like something that only one person wanted, and I don't think adding a huge amount of extra 3rd-party code is a great idea anyway.  I haven't really reviewed the umpteen commits either, but I got to a big chunk of HTML being assembled with StringBuilder.Append and rather gave up.

Like you, if a captcha is useful, I'd prefer to see a clean implementation that used a library.

I don't think we should add any new functionality unless/until we can get 6.0.0 out properly.

RedX2501:
Hey, sorry for the late reply.

I agree with both of you. We should use a library or a third party service.

larshg:
@68681395 We would like a better implementation of this, before considering it to be merged. Please open a new PR with a proper implementation.

68681395:
well, I will open a new PR when i maybe package the captha function as a individual library.

PR #613, opened 2016-11-15
fix routing and blob display bugs

lkho:


Bonobo.Git.Server/Views/Repository/Blob.cshtml
```diff
@@ -13,15 +13,15 @@
...
+                <a href="@Url.RouteUrl("RepositoryBlame", new { id = ViewBag.ID, encodedName = PathEncoder.Encode(Model.TreeName), encodedPath = PathEncoder.Encode(Model.Path, allowSlash: true) })"><i class="fa fa-comments"></i> @Resources.Repository_Tree_Blame</a>
+                <a href="@Url.RouteUrl("RepositoryHistory", new { id = ViewBag.ID, encodedName = PathEncoder.Encode(Model.TreeName), encodedPath = PathEncoder.Encode(Model.Path, allowSlash: true) })"><i class="fa fa-history"></i> @Resources.Repository_Tree_History</a>
+                <a href="@Url.RouteUrl("RepositoryRaw", new { id = ViewBag.ID, encodedName = PathEncoder.Encode(Model.TreeName), encodedPath = PathEncoder.Encode(Model.Path, allowSlash: true), display = true })"><i class="fa fa-file-text"></i> @Resources.Repository_Tree_RawDisplay</a>
+                <a href="@Url.RouteUrl("RepositoryRaw", new { id = ViewBag.ID, encodedName = PathEncoder.Encode(Model.TreeName), encodedPath = PathEncoder.Encode(Model.Path, allowSlash: true) })"><i class="fa fa-download"></i> @Resources.Repository_Tree_Download</a>
             </div>
```
willdean: Seems like this isn't really fixing a bug - it's just renaming a bunch of actions to include the controller name within the action name - what was the motivation here?

Bonobo.Git.Server/Views/Repository/Blob.cshtml
```diff
@@ -13,15 +13,15 @@
...
         }
         @if (Model.IsImage)
         {
-            <img class="fileImage" alt="@Model.Name" src="@Url.Action("Raw", new { encodedPath = PathEncoder.Encode(Model.Path, allowSlash: true) })" />
+            <img class="fileImage" alt="@Model.Name" src="@Url.RouteUrl("RepositoryRaw", new { id = ViewBag.ID, encodedName = PathEncoder.Encode(Model.TreeName), encodedPath = PathEncoder.Encode(Model.Path, allowSlash: true) })" />
         }
```
willdean: Nice one, thanks - I'll commit this outside this PR while we get the rest sorted.

willdean:
Thanks very much for this stuff - there are some problems with just merging as it is however:

* It seems to be a mix of bug fixes and stylistic changes
* Changing the URLs breaks the automated test
* It's not always clear what the bugs are that are being fixed are - 68c1cc3 is definitely a real fix to a real problem, but I couldn't repro it here on a quick test with a .png file - could you expand on what this fixes?

PR #617, opened 2016-11-24
Adding new entries for translation pt-BR and Removing CultureInfo.InstalledUICulture 

holiiveira:
About Removal CultureInfo.InstalledUICulture check issue 616.


willdean:
Hi - thanks very much for this - would you mind removing the last commit from this, so that we can just merge the language changes separately?   There are other places where the InstalledUICulture is (mis)used, and I'd like to deal with them in one go.  You may need to create a new PR with just the two translation commits in it.

PR #626, opened 2016-11-30
Allow to use c#6 with old visual studio

latop2604:
Now, we can use `string interpolation`, `nameof `and `other c#6` features with back compatibility with old Visual Studio.
The Microsoft.Net.Compilers nuget package embed the roslyn compiler and replace the default one.


willdean:
This is very cool - I didn't know it was possible.

I'm not really sure if it would be a problem to move to VS2015+ anyway .

PR #629, opened 2016-12-06
Compatibility with IE version.

holiiveira:
I was having compatibility issues with more current versions of IE, this configuration solved my problem.

### DIFF
--- a/Bonobo.Git.Server/web.config
+++ b/Bonobo.Git.Server/web.config
@@ -76,6 +76,12 @@
         </hiddenSegments>
       </requestFiltering>
     </security>
+    <httpProtocol>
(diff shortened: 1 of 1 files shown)


RedX2501:
What kind of problems were you having?

holiiveira:
No problem about the operation, only problem with layout (buttons that do not appear), maybe some incompatibility of javascript, I do not know for sure.

willdean:
I normally just add this into the head of the layout page rather than adding more stuff to web.config.  I do think we should have it though, as without it some versions of IE lag a long way behind current browsers.

holiiveira:
I understand, your solution is cool.

Issue #562, opened 2016-05-09
AD plugin not working for domain forest


mr-miles:
Hi,

I have set up bonobo git server at my workplace and it is all looking good.  We have a weird AD setup with multiple domains, and I am finding that users from the non-default domain are not able to log in.  I think the problem is that they are not being created as a user in the backend.

I think I've traced this back to UpdateUsers.UpdateUsers.

The code gets the group and enumerates over the members.  However, for each member it then attempts to retrieve the principal from the default domain.  In my case this returns null because the user needs to be retrieved from the domain specified in their UserPrincipalName.

For example, if the members group is DOMAIN1\GitGroup and the problem user is DOMAIN2\ProblemUser, then Problemuser can only be retrieved from DOMAIN2.

Are you open to a pull request?

Thanks

Miles

willdean:
Yes, we would definitely consider a pull-request.   We generally struggle to test/maintain the AD stuff at the moment, so good-quality AD work will probably be well-received.

If you're testing the 5.2 rather than with the code from the master branch, you should definitely work with the master code, because there have been huge changes since 5.x.  

I am not looking at the code right now, but ISTR there are already some places in the code where the domain is extracted from the UPN and used instead of the default domain - if you've found a place where this isn't happening, then perhaps you could try and find a consistent solution.

larshg:
@mr-miles do you have multiple servers as well with different DNS names? And then a trust between these servers?

larshg:
@mr-miles have you tried the recent releases?

larshg:
Closing this due to no response. If the users from the second domain is added to the first domains usergroup - it should be possible that the users is found in the global catalogs. But I'm far from certain about this.

PR #637, opened 2017-01-25
Use default domain as it might be different from mail domain.

### LINKED ISSUE #602: AD Install loses repo settings in upgrade from 5.2 to 6.0 pre-release
In #587 @Trapulo posted 

> @willdean tried to update to last build. Lost every repo details (users permissions, logos, and other details) but all seems ok. Tried a little sync and commit from Visual Studio and all seems ok. Is there some specific trouble may I search?

@Trapulo, assuming you have a backup copy of your old 5.2 install, could you see if you have an "ADBackend" directory in app_data for both 5.2 and 6.0 installs, and if you can see any obvious difference between the files in the "Repos" subdirectory in both installs.

I don't think it should have lost all your repo settings on that upgrade.

larshg:
Alternatively implement to test both. Might fix #602

### DIFF
--- a/Bonobo.Git.Server/Content/Site.css
+++ b/Bonobo.Git.Server/Content/Site.css
@@ -1,4 +1,4 @@
-﻿html, button, input, select, textarea, .pure-g [class *= "pure-u"], .pure-g-r [class *= "pure-u"] { font-family: OpenSans, sans-serif; }
+html, button, input, select, textarea, .pure-g [class *= "pure-u"], .pure-g-r [class *= "pure-u"] { font-family: OpenSans, sans-serif; }
 
 html { font-size: 16px; color: #333; text-rendering: optimizeLegibility; }
 html, body { margin: 0; padding: 0; height: 100%; min-width: 1100px;}
@@ -155,7 +155,7 @@ Tags .tag a { background-color: rgba(148, 190, 214, 1); border-radius: 5px; colo
 .commit .metadata p { margin: 0; font-size: 0.8em; font-family: Consolas, monospace;}
 .commit .metadata h4 { margin-bottom: 0.5em;}
 .commit .metadata { text-align: right;}
-.commit h2 {width: 780px; overflow: hidden; text-overflow: ellipsis; white-space: nowrap;}
+.commit h2 {width: 780px; overflow: hidden; text-overflow: ellipsis; white-space: nowrap; margin-bottom: 0; position:relative;}
 .commit h2 a:hover { text-decoration: none; color: #777;}
 .commit { border-bottom: 1px solid #a6a6a6; padding-bottom: 3px; }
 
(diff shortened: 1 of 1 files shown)


willdean:
I am really weak on this stuff, and can't tell what other problems this creates - I like the option you mentioned in #602 of trying for the domain in the email address and falling back to the default if that doesn't work - that feels safer than just using the default-domain.  I have a feeling that the whole business of pulling domains out of email addresses started because there were people who couldn't just use the default domain.

larshg:
Ahh okay - I'll eventually try to make the required changes then.

larshg:
@willdean I have made the required changes. @RedX2501 can you have a look too?

willdean:
Looks good to me - I can't test this here, but I think the approach looks OK.

willdean:
@larshg Are you happy this works OK on your setup for importing users that have both the default domain and a different domain?  If so then I think we should merge it and then push for a 6.0-beta release.  @RedX2501  is busy on other things at the moment and I don't think he can easily test AD stuff anyway.

larshg:
Well, yes and no... It seems to work, but if one extracts the domain from the UPN (email) name and its different from the domain controller, it takes a long time to timeout and try the default domain... But I can't really come up with a better solution - at least for now - so lets go ahead and merge, to get some feedback?

larshg:
I have just read more about AD, domains and UPN suffixes... For what I understand, the UPN suffixes is rarely the same as the domain name of the AD server and hence I don't think we should extract the suffix from logon name etc.

Either we should use the default domain from settings - or we should create a principal context with no domain specified and it will try to connect whatever the webserver domain is part of?

One explanation of additional suffixes can be read [here](http://www.tutorialspoint.com/articles/adding-alternate-upn-suffix-to-active-directory-domain).

Naming of AD servers FQDN is dicussed [here](http://serverfault.com/questions/76715/windows-active-directory-naming-best-practices)  - ie. ad1.domain.com.

I don't think that if we extract domain.com from a logon name that the AD server would respond on this - because it should be on a subdomain...

larshg:
I guess the original problem was reported by #562 from @mr-miles.

willdean:
@larshg If we shouldn't be pulling domain names out of user names at all, does that mean that routines like `GitAuthorizeAttribute.IsWindowsUserAuthorized` are mistaken?

That uses a helper called `StripDomain` (it's mis-named, it should really be called `ExtractDomain`) to get a domain out of user@domain and then builds a PrincipalContext from that. 

I too am concerned about long delays in the import, that something in IIS will then interrupt the import and cause all sorts of impossible to debug problems with partial imports.

Maybe we should just try to make things work properly using the 'Default Domain' for a 6.0-beta, and then look careful at situations where this doesn't work properly for people?   (I'm sorry as I know that's where you started with this PR)

larshg:
Well, that is what I get out of all the reading :) - also, when I create the PrincalContext with the adress of our domain server my user is validated using the full UPN name - whereas the current `StripDomain` does remove the suffix of the UPN ie. @domain.com - but this shouldn't be necessary.

An alternative to the multi domain/server problem could be to change the DefaultDomain to a list of AD servers, which is then iterated. As long as the server is connected, the verification of a valid user is quite fast, so it shouldn't take long to validate a couple of servers? But again... if there are servers not up and running, it can take a long time to timeout the servers that are not available. Both on upgrade, but also when verifying all the git operations etc.

I'm still trying to read up on how to handle trusted domains from one domain to another in a multi domain/forest setup :)

willdean:
The project so needs an AD expert - it's great that you've selected yourself  :-) 👍

larshg:
A Microsoft [article](https://msdn.microsoft.com/en-us/library/ff650307.aspx) that describes what I suggested about multiple servers :)

willdean:
@larshg Doesn't that article describe exactly what we have been doing, which is obtaining a domain name by splitting it out of a UPN?

larshg:
Unfortunately yes - it does... :/

larshg:
[This](https://www.codeproject.com/Articles/18102/Howto-Almost-Everything-In-Active-Directory-via-C) article might be good read. I can see that we can access the forest and get first the root domain and any subdomain / servers and iterate all, until a user is found etc.

I'll try read up on it during the week.. Lets wait a bit with merging!

PR #644, opened 2017-02-10
[RFT]Changed to iterate through the domains in the forest.

larshg:
This doesn't use the defaultdomain from settings at all atm.

### DIFF
--- a/Bonobo.Git.Server/Helpers/ADHelper.cs
+++ b/Bonobo.Git.Server/Helpers/ADHelper.cs
@@ -1,28 +1,35 @@
 ﻿using Bonobo.Git.Server.Configuration;
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.DirectoryServices.AccountManagement;
 using System.DirectoryServices.ActiveDirectory;
-using System.Linq;
-using System.Web;
 
 namespace Bonobo.Git.Server.Helpers
 {
     public static class ADHelper
     {
-        public static bool ValidateUser(string parsedDomain, string username, string password)
+        /// <summary>
+        /// Validates the user against Active directory - will try the domain that are part of the username if present
+        /// Alternatively it tries all domains in the current forest.
+        /// </summary>
+        /// <param name="username">Username with or without domain</param>
+        /// <param name="password"></param>
+        /// <returns>True on successfull validation</returns>
+        public static bool ValidateUser(string username, string password)
         {
+            var parsedDomain = username.GetDomain();
+            string strippedUsername = username.StripDomain();
+
             Domain matchedDomain = GetDomain(parsedDomain);
-            // If a domain was present in the supplied username, try to find this first at validate against it.
+            // If a domain was present in the supplied username, try to find this first and validate against it.
             if(matchedDomain != null)
             {
-                return ValidateUser(matchedDomain, username, password);
+                return ValidateUser(matchedDomain, strippedUsername, password);
             }
             // Else try all domains in the current forest.
             foreach (Domain domain in Forest.GetCurrentForest().Domains)
             {
-                if (ValidateUser(matchedDomain, username, password))
+                if (ValidateUser(matchedDomain, strippedUsername, password))
                     return true;
             }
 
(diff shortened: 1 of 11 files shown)


willdean:
@larshg Awesome!  I don't really understand the details, but does it work OK for your setup?

On thing I think that's important about the migration is to keep going if there's a problem with one user/repo/team/whatever.   That way people might lose a single (or a few) problematic item during a migration, but they won't lose it all.  If nothing else that will make supporting them easier, because the symptoms will be better isolated.   

Do you think this meets that requirement?  I see quite a lot of try-catch with logged exceptions, so we're probably pretty good there...

larshg:
I think it needs a few more additions [here](https://github.com/jakubgarfield/Bonobo-Git-Server/blob/09014c8732bbb6628e2972fa75b423b98eba652f/Bonobo.Git.Server/Data/Update/ADBackend/UpdateADBackend.cs#L19-L40). Because if either of the 4 methods fail, we won't really know which one is buggy as the errors are not catch separately.
And as I review my on commit I can see that there is room for general improvements 🗡 

But I pushed it for now to get more review it etc.

cbaerikebc:
Hmm, this may be problematic:
`foreach (Domain domain in Forest.GetCurrentForest().Domains)
+            {
+                try
+                {
+                    using (PrincipalContext pc = new PrincipalContext(ContextType.Domain, domain.Name))
+                    {
+                        if (pc.ValidateCredentials(username, password, ContextOptions.Negotiate))
+                            return true;
+                    }
+                }`

We have multiple domains (in the same forest), but usernames are not unique across domains (e.g. there could be a user abc in 3 different domains, with different passwords etc.). Users get locked out after 3-5 invalid password attempts (depending on the domain). In the following scenario, users will get locked out of their accounts with the above method:
- Let's assume we have domains foo.local, bar.com, and xyz.net. They all have a user abc with different passwords.
- The user who's trying to authenticate is part of the xyz.net domain, but the above code may try foo and bar first.
- User xyz\abc will always authenticate successfully, but in the background each authentication attempt will increase the BadPassword counter for the other 2 users until they're locked out.

Could you pass the user-entered domain name (the part that is stripped away earlier) to this function as well, and only try to authenticate the user in domains that contain the user-entered domain name? E.g. the user authenticates as xyz\abc, so xyz is passed into this function as the domain. It should then try to authenticate the user against xyz.net, but not foo.local or bar.com.

larshg:
Thanks for the feedback @cbaerikebc. Indeed this was also something I thought about, but I have the case that I have added a additional UPN name - and other could have this too - when the login with UPN name ie. myname@mydomain.com. This UPN suffix would the not match a AD server and take a lot of time trying to connect.

But a solution could be to pass in the domain and try to see if it match any of the domains in the forest and if it does, connect to that one first. If there is no match, then try the domains found.

cbaerikebc:
Yes, that should work.

larshg:
@cbaerikebc and @willdean  I have updated it a bit.

There is only a small issue if one have several domains with same name, but various endings . ie domain.net, domain.com and only supply the username in SamAccount form ie. domain1\\user.

But I guess its really complicated to cover all scenarios?

cbaerikebc:
I think that's alright. In that case, the supplied username would no longer be unique. The only way to fix that would be to define a search order, e.g. in the web.config.

Bonobo.Git.Server/Helpers/ADHelper.cs
```diff
@@ -11,30 +11,53 @@ namespace Bonobo.Git.Server.Helpers
...
 
+        private static Domain GetDomain(string parsedDomain)
+        {
+            foreach (Domain domain in Forest.GetCurrentForest().Domains)
+            {
+                if(domain.Name.Contains(parsedDomain))
+                    return domain;
+            }
+            return null;
+        }
```
willdean: Is everywhere that calls this OK with a possible null return?   Would an empty string be better?  (I don't know, just asking!)

Bonobo.Git.Server/Security/ADMembershipService.cs
```diff
@@ -32,9 +32,12 @@ public ValidationResult ValidateUser(string username, string password)
             if (String.IsNullOrEmpty(username)) throw new ArgumentException("Value cannot be null or empty", "username");
             if (String.IsNullOrEmpty(password)) throw new ArgumentException("Value cannot be null or empty", "password");
 
+            var parsedDomain = username.GetDomain();
+            var strippedUsername = username.StripDomain();
```
willdean: Are we OK with nulls here?  Is that possible/is it handled OK?

willdean:
I don't feel qualified to review the AD stuff properly, but it looks like clean code to me, as long as you're happy that callers and methods are clear about null-returns.  I don't understand the appveyor test failure which doesn't feel like anything related to this.

If you and @cbaerikebc are happy with this, then I think we should merge it and release 6.0 (assuming the tests pass).

Bonobo.Git.Server/Helpers/ADHelper.cs
larshg: It has to return an object of type Domain - not a string :) and yes, it should handle a possible null where it is called.

larshg:
Should we remove/disable the Application.DefaultDomain - as it is no longer used?

willdean:
LGTM - go for it

larshg:
Just pushed minor updates after trying out the artifact on my prod environment.

willdean:
@larshg Do you think we can merge this now?

larshg:
Well, I guess it would be nice if those that actually have multiple domains can test it first?

willdean:
@cbaerikebc I know it's a big favour to ask, but is there any possibility you can try this version with your multi-domain setup?

larshg:
@mrmiles can you test out [this](https://ci.appveyor.com/project/jakubgarfield/bonobo-git-server/build/507/artifacts) aswell?

cbaerikebc:
@willdean This has been on my list to test since the fix was submitted; the issue is finding the time to do so. I will try, but at the moment can't promise that I can test this before April.

willdean:
@cbaerikebc Thanks!  No worries if you can't do it, we understand...

larshg:
Should we just merge this to get 6.0 out - I think nevertheless that its a consolidation of previous AD code that was scattered all over and a bit unorganized? If it doesn't work its properly easier to go back to the "old" way or continue on from here?

willdean:
@larshg Yes, I think we should merge it, and release 6.0 and see what happens - if there's a shitstorm and we have to do a 6.1 in a hurry then that's just bad luck, but otherwise we're just paralysed.

KY4U:
Dang, after deploying this code, now I'm getting a 500 server error after login in.

larshg:
Please open a new issue and explain when you get the error - and with this code I assume you mean using avtive directory and Windows authentication ?

PR #651, opened 2017-03-06
Updated CommonMark.NET to veresion 0.15.1

larshg:
Tested it by visually comparing the file supplied from #570.


larshg:
I don't think the failed tests have anything to do with the CommonMark...

PR #652, opened 2017-03-09
Updated the rest of the nuget packages.

larshg:
Updated Newtonsoft.JSon
Updated System.IdentityModel
This required .net 4.5.2> so I took 4.6
Updated SQLite
Updated SpecsFor
Updated Selenium
Updated Microsoft.Net.Compilers
Updated DotNetZip
Updated WebActivatorEx

I'm not entirely sure where the updates to the designer file is from?


larshg:
@RedX2501 I have some problem getting the repository created in the integration unit test - how can I debug this? I tried adding more trace logs, but nothing shows up in the log file?

larshg:
It was the missing dlls to git... 

@willdean do you think there is any problems in merging this?

willdean:
I guess if all the tests pass we should be OK to merge - I can't see what else we'd wait for...

PR #658, opened 2017-03-27
Update Global.asax.cs

Rakkoc:
ToLower() to ToLowerInvariant() because another lang can not translate "BinaryGUID=false" to "binaryguid=false", ToLower translate translate to "binaryguıd=false".


willdean:
Thanks for this, now merged

PR #660, opened 2017-03-30
add compare difference between 2 commits and filter with a filepath

lang4net:
features:
1. add compare difference between 2 commits and filter with a filepath
2. add back to link on patch panel


changelog.md
```diff
@@ -3,6 +3,15 @@ title: Changelog
...
 ---
+## New Version
+
+### Features
+
+* compare the difference between 2 commit filter with a filepath
+* commit view: go to change, add back to link
+
+* 增加Compare功能：比较两次提交之间的所有变更详情，可以通过文件名或文件夹过滤。
+* 优化Commit导航：patch窗口上的图标可以快速返回页面顶部的changes导航链接。
```
willdean: There seem to be Chinese characters here - I think the changelog.md file needs to stick to English

willdean:
Many thanks for this - I'm sorry but it's not really clear to me what's being proposed here.  

Is this two, unrelated, features?  If so, can you put them in as two PRs?

lang4net:
I'm sorry,it seems that there are too many changes.
i will keep the PR clear,pull it later

changelog.md
lang4net: yes, the above is english, the following is the chinese translation.

PR #670, opened 2017-04-10
Added missing french resource keys

franckbour:
Correction on some sentences and missing translations added for french language


willdean:
Tres bon!  Merci beaucoup... (I'm too English to know how to type an e-acute, je sui désolé...)

Issue #668, opened 2017-04-09
Windows authentication with internal membership service failed


aPSopt:
For details see [http://stackoverflow.com/questions/43293352/bonobo-git-server-authentication-failed-why](http://stackoverflow.com/questions/43293352/bonobo-git-server-authentication-failed-why)
I download and install Bonobo Git Server (6.0.0.532). I add the App_Data folder permissions and I have IIS authentication settings as described in the installation notes - [https://bonobogitserver.com/install/](https://bonobogitserver.com/install/)
I create the user (login: qqq, pass: qqq) and try to clone repository with this user.
But I see authentication failed error all time.

I look in the config and see, that there are turn on internal membership service - it is default behaviour of the bonobo git server. But it is not works. So this is the bug of the bonobo git server or bad documentation.

brianhv:
I'm also experiencing this behavior.

I initially attempted to upgrade from 3.4, and cloning failed. Thinking it was an issue with the upgrade, I then installed 6.0 in a fresh directory and just enabled Windows authentication (via web.config, not via IIS authentication). Cloning still failed no matter how I specified the username.

I went one step further and tried with a completely stock 6.0, with the default cookie authentication. I created a new repository and a new user, giving the new user access to the repository. The new user successfully logs in through the web interface, but fails to clone the repository with an uninformative 401.

This is currently running on Windows Server 2008, IIS 7.5.

brianhv:
This appears to have similarities to #567, which is somewhat old but unfortunately lacking a solution.

willdean:
@brianhv Assuming you're using a normal sort of command line Git client, when you do:

`git clone <git repo>`, do you get asked for credentials?  Or does some credentials helper supply them without asking you?

If it's the latter, could you try disabling that (it's probably configured in your global .gitconfig file in your home directory) and see what happens?

brianhv:
I do get asked for credentials. I've tried providing the username as `brianhv` (as I do through the web), and as `domain\brianhv` (the way the user is displayed in the web interface). The result in both cases has been the terse `fatal:  Authentication failed` message as shown in @aPSopt's stackoverflow post.

At this point I've downgraded back to 3.4 and it's working for now. And it turns out that this time, 6.0 with cookie auth is working; it's only windows authentication mode that fails from the git client. (Which for me is a macOS command-line client.)

[...]

PR #671, opened 2017-04-11
Add Serilog logging

### LINKED ISSUE #490: Logging?
Bonobo currently has fairly simple logging, just writing exceptions to a file.

I think some of the problems people report issues on could be more easily debugged if there was better logging in the application - for example lots of the interaction with AD could be logged at various levels, which could be very helpful when trying to diagnose problems, and there are lots of other situations where we have some indication that something's going wrong, and logging a warning could be helpful.

What do people think about adding a logging framework and starting to make use of it?
Personally I'd probably use NLog or Serilog, but I'm not fussed if anyone else has strong feelings about something else.

willdean:
Fixes #490 

Second commit also reverts recent removal of windows auth code out of GitAuth attribute - have a look at the history on GitAuthorizeAttribute to see the same thing happening previously.

Am trying to make some progress on issues like #668

### DIFF
--- a/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
+++ b/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
@@ -98,21 +98,75 @@ namespace Bonobo.Git.Server
 
             if (!String.IsNullOrEmpty(username) && !String.IsNullOrEmpty(password))
             {
-                Log.Information("GitAuth: Going to membership service for user {username}", username);
-
-                if (MembershipService.ValidateUser(username, password) == ValidationResult.Success)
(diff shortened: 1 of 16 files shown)


willdean:
@larshg Not sure if you have a chance to look at this, but the second commit backs out a change to GitAuthorizeAttribute which I originally merged from your PR.  

If you look at the history of this file, we've mistakenly passed Windows auth to the membership provider before, but that's not right.  It's probably broken in a different way in this PR, but it might help people who are currently struggling with using Windows auth with EF membership.

larshg:
Can you maybe update to use the ADHelper class to properly iterate multiple domains etc.

And add the additional logging here if you haven't already done this :-)

willdean:
I'm closing this because I don't like it being a mixture of logging and auth fixes - I'm going to add separate new PRs

PR #674, opened 2017-04-12
Win auth fixes

willdean:
Fix the Windows Auth with EF Membership scenario - see #668 

@larshg - could you have a very quick look - it does seem that your ADHelper does all the heavy lifting and there was nothing much left to do in GitAuthAttr, which is great if it's true.

@brianhv Would you mind trying this one when it's built, please?

### DIFF
--- a/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
+++ b/Bonobo.Git.Server/Attributes/GitAuthorizeAttribute.cs
@@ -4,7 +4,6 @@ using System.Security.Claims;
 using System.Text;
 using System.Web;
 using System.Web.Mvc;
-using System.DirectoryServices.AccountManagement;
 using Bonobo.Git.Server.Data;
 using Bonobo.Git.Server.Security;
 using Microsoft.Practices.Unity;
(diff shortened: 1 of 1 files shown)


larshg:
All logging is also in this one :-) ?

willdean:
Yes, but they're two branches, and once I merge the logging one separately this one will look more sensible...

larshg:
Ah, for what I can validate from my phone, it seems fine :-)

larshg:
Ah, for what I can validate from my phone, it seems fine :-)

brianhv:
Tested. Works for me!

larshg:
@willdean can you "release" the 6.1? or should we do some more logging in AD sync etc for better support in this area?

willdean:
@larshg I did it last week - I think maybe I forgot to tag this repo, though, so I'll do that now.  It's up at https://bonobogitserver.com/

larshg:
ahh, I just lookhed [here](https://github.com/jakubgarfield/Bonobo-Git-Server/releases) :)

willdean:
Yes, thanks for the reminder!  You can look again now...

PR #677, opened 2017-04-20
Fix translator language pt-BR removed redundant.

afonsoft:
modified file: 

Bonobo.Git.Server/App_GlobalResources/Resources.pt-BR.resx

Fix translator language pt-BR removed redundant.

For layout of buttons on /Repository/Index

### DIFF
--- a/Bonobo.Git.Server.sln
+++ b/Bonobo.Git.Server.sln
@@ -1,11 +1,9 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio 2013
+VisualStudioVersion = 12.0.40629.0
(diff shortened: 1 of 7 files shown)


willdean:
Thanks for this - help with the translations is always great.

Could you please re-do this commit with just the translation changes in, not all the changes to the project files, etc?  Thanks.

willdean:
Getting better, but there are still two unrelated new files in there...

willdean:
OK - thanks - you now have three commits on this branch, which isn't really what we want, because we can't merge them like this.   (I could mess around for ages sorting it out be I'd rather not)

Rather than push everything to 'master' on your own repo, you would be better to make a branch on your repo, commit the changes to that branch, and then let GitHub make a PR from that branch.

Thanks for the help - I am not trying to be discouraging, but it's good if you can make PRs in a form which people find easy to merge.

PR #680, opened 2017-04-22
Added additional logging and removed old Trace log.

larshg:
Added method to dump all groups in domain.

PR #690, opened 2017-05-09
More logging

willdean:
Remove all remaining use of TraceListener for logging


willdean:
@larshg Sorry - this one probably makes #680 even harder to merge - perhaps if you think this one's OK then we'll take this first and then the AD group logging stuff from 680?

Issue #687, opened 2017-05-07
Non-admin user can modify info in User page and in Team page


mrloddo:
Hi all,
I've already tested a recent version of Bonobo Server. Now the AD login works fine using Windows auth and internal membership.
I found now a non-admin user is able to change, delete all users and team info. The account is able to elevate itself to administrator role.

Marco

willdean:
@mrloddo Thanks for this - are you saying that the account could delete all users before you elevated it, or that somehow you elevated it first?

Could you perhaps give us a set of steps to reproduce this, starting with when you created the user in the internal membership system?

larshg:
I have verified this - a user which should not have admin privileges, does in fact have these. I'll try narrow it down where and why this happens.

larshg:
The error occurs only on first login - since the user is not found in the database when the claims is queried [here](https://github.com/jakubgarfield/Bonobo-Git-Server/blob/master/Bonobo.Git.Server/Owin/WindowsAuthenticationHandler.cs#L50) - the user is looked up on the AD server and a new ClaimsIdentity is [created](https://github.com/jakubgarfield/Bonobo-Git-Server/blob/master/Bonobo.Git.Server/Owin/WindowsAuthenticationHandler.cs#L73), with info from AD and double [admin](https://github.com/jakubgarfield/Bonobo-Git-Server/blob/master/Bonobo.Git.Server/Owin/WindowsAuthenticationHandler.cs#L89-L90) claims.

This is then set in the authentication [ticket](https://github.com/jakubgarfield/Bonobo-Git-Server/blob/master/Bonobo.Git.Server/Owin/WindowsAuthenticationHandler.cs#L99) - and therefore the use has admin rights immediately.

However the user is only having the role assigned if the flag or there are only 2 users in the database [here](https://github.com/jakubgarfield/Bonobo-Git-Server/blob/master/Bonobo.Git.Server/Controllers/AccountController.cs#L204)

Hence on subsequent logins the users claims will queried from the database, where the user doesn't have admin rights.

If we remove the two lines adding the Administrator rights, the user won't have those when redirected to the index page, since the Authentication ticket is not updated with the newly acquired adminrights created in the ADCreateUser.

I don't know if it is possible to add the claim to the ticket or if we should clear it and force the user to logon again (which should happen automatically if the browser trust the server and uses NTLM/windows logon) ?

@willdean do you have any good ideas?

willdean:
[...]

PR #691, opened 2017-05-10
Don't allow unknown AD users to be admins before ADCreateUser

willdean:
Potential fix for #687

### DIFF
--- a/Bonobo.Git.Server/Controllers/AccountController.cs
+++ b/Bonobo.Git.Server/Controllers/AccountController.cs
@@ -181,7 +181,9 @@ namespace Bonobo.Git.Server.Controllers
 
         public ActionResult CreateADUser()
         {
-            if ((!Request.IsAuthenticated) || !(MembershipService is EFMembershipService))
+            var efms = MembershipService as EFMembershipService;
+            
+            if ((!Request.IsAuthenticated) || efms == null)
             {
                 Log.Warning("CreateADUser: can't run IsAuth: {IsAuth}, MemServ {MemServ}", 
                     Request.IsAuthenticated,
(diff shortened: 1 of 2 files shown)


willdean:
@larshg It's [this line](https://github.com/jakubgarfield/Bonobo-Git-Server/pull/691/commits/bdec7efff9e79efda863b37e9bf17cd72ff12576#diff-05c1790487572d68236e1424d03dd7c2R208) which I think should make the current user into an admin.

larshg:
I'll try it out tomorrow :-) but it looks Loke it should work!

willdean:
@larshg Thanks!  What timezone are you in?  (I am in the UK, so UTC+1 in the summer)

larshg:
For some odd reason, the added claim is not present after the redirect. Somehow it is not properly written to the cookie? The ClaimsIdentity is created with string litteral(SignInAsAuthenticationType) specifying "Cookies". 
I guess thats how we avoid having to reauthenticate against the AD all the time?

willdean:
We might need to sign in again after adding the claim?  Have a look at this;

http://stackoverflow.com/questions/20383955/how-to-add-claims-in-asp-net-identity

Perhaps if we do the 'signin' bit again after adding the claim?

willdean:
Or the accepted answer on [here](http://stackoverflow.com/questions/24587414/how-to-update-a-claim-in-asp-net-identity) might be useful?

larshg:
Yes, the last suggestion works.

I have made some changes, but for some reason github is rejecting my pushes atm.

larshg:
On a site note... Isn't it a security risk saving rights in Cookie - then as a long as this cookie lives, the user can have admin roles etc. I tried for the fun of it to promote a second user to admin, but I had to relog and clear cookies for it to get the new roles from the DB.

Also the other way around, when I removed the admin rights of the first account, it first disappeared after logging out and clearing the cookies.

I know its a short period of time and properly a rare case... but sometimes you could have a crazy colleague :D

larshg:
Uhh and I'm in DK, so its the same, UTC +1 :)

willdean:
>I have made some changes, but for some reason github is rejecting my pushes atm
Are you trying to push to my branch - that's probably the problem - I'll give you access, hang-on...

willdean:
@larshg You should have had a collaboration invite for my fork from GH?

larshg:
@willdean - I was trying to push to my own repository :-) 
But I'm busy this weekend, but maybe I get a spare minute to try to push again :-)

PR #703, opened 2017-06-17
Enable background server garbage collection

tweber-cs:
https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/fundamentals#background_server_garbage_collection

### DIFF
--- a/Bonobo.Git.Server/web.config
+++ b/Bonobo.Git.Server/web.config
@@ -79,6 +79,7 @@
     </staticContent>
   </system.webServer>
   <runtime>
+    <gcServer enabled="true"/>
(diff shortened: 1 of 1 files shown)


willdean:
According to (my reading of) that article this is the the default anyway, isn't it?

tweber-cs:
Based on these docs the default is enabled="false"

https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/file-schema/runtime/gcserver-element

Value | Description
------------ | -------------
false 	| Does not run server garbage collection. **This is the default.**
true 	| Runs server garbage collection.

PR #705, opened 2017-06-21
Allow Custom CSS URLs

damienhoneyford:
To allow for further customisation beyond the Logo, Title and Footer Message, include a CSS URL that can be used to style the Bonobo instance without needing to modify the inbuilt stylesheet, which can be easily overwritten during an upgrade.

### DIFF
--- a/Bonobo.Git.Server/Views/Shared/_Layout.cshtml
+++ b/Bonobo.Git.Server/Views/Shared/_Layout.cshtml
@@ -5,6 +5,10 @@
     <link rel="shortcut icon" href="~/favicon.ico?v=2" type="image/x-icon" />
     <title>@ViewBag.Title :: @UserConfiguration.Current.GetSiteTitle()</title>
     @Styles.Render("~/Content/bundledCss")
+    @if (UserConfiguration.Current.HasCustomSiteCss)
+    {
+        <link rel="stylesheet" href="@Url.Content(@UserConfiguration.Current.SiteCssUrl)@(string.Format("?_={0}", DateTime.Now.Ticks))" type="text/css" />
+    }
     @if (UserConfiguration.Current.HasCustomSiteLogo)
     {
     <style>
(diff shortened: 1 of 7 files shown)


Bonobo.Git.Server/Views/Shared/_Layout.cshtml
```diff
@@ -5,6 +5,10 @@
     <link rel="shortcut icon" href="~/favicon.ico?v=2" type="image/x-icon" />
     <title>@ViewBag.Title :: @UserConfiguration.Current.GetSiteTitle()</title>
     @Styles.Render("~/Content/bundledCss")
+    @if (UserConfiguration.Current.HasCustomSiteCss)
+    {
+        <link rel="stylesheet" href="@Url.Content(@UserConfiguration.Current.SiteCssUrl)@(string.Format("?_={0}", DateTime.Now.Ticks))" type="text/css" />
```
willdean: Style sheet will *never* be cached - was that the intention? Is that a good thing?

damienhoneyford: Yes, that was the intention.  In all likelihood the custom stylesheet would be very small, so not caching it will add a small amount of overhead to each request, but guarantee that the latest version is always served.

willdean: OK, cool - do we need three `@` signs in there or could it actually just be one `@(...)` expression?

damienhoneyford: Yes, it could definitely be tidier -- I've committed a change to my forked repo which tidies that code up.

willdean:
Thanks for the updates.  I'm inclined to merge this, as it seems pretty harmless to people who don't like it / want it.

The only possible downside is that people who've written custom CSS files might object if we then changed class names or IDs during an update, but that's probably a minor concern.

Not sure if @larshg is around for a second opinion...

PR #710, opened 2017-07-05
Browsing / acessing public/anonymous repositories w/o credentials

jeffgaroutte:
https://github.com/jakubgarfield/Bonobo-Git-Server/issues/77

I updated the WebAuthorizeRepositoryAttribute and the RepositoryController to allow anonymous access to repositories that have it turned on.

I also updated the repository index view to remove the delete, add and personal repository URL


Bonobo.Git.Server/Views/Repository/Index.cshtml
```diff
@@ -166,9 +166,9 @@
...
-                    columns.Add(permitted.Column(format: (item) => Html.ActionLink(" ", "Edit", new { id = item.Id }, new { @class = "edit", title = Resources.Grid_Edit }), style: "action"));
-                    columns.Add(permitted.Column(format: (item) => Html.ActionLink(" ", "Delete", new { id = item.Id }, new { @class = "delete", title = Resources.Grid_Delete }), style: "action"));
+                    columns.Add(permitted.Column(format: (item) => Html.Raw(String.IsNullOrWhiteSpace(item.PersonalGitUrl) ? "&nbsp" :"<button class=\"pure-button pure-button-primary copy-to-clip-button\" title=\"" + Resources.Repository_Personal_Git_Url_Copy + "\" data-clipboard-text=\"" + Html.Encode(item.PersonalGitUrl) + "\"><i class=\"fa fa-user\"></i><i class=\"fa fa-copy\"></i></button>"), style: "action"));
+                    columns.Add(permitted.Column(format: (item) => User.IsInRole(Definitions.Roles.Administrator) ? Html.ActionLink(" ", "Edit", new { id = item.Id }, new { @class = "edit", title = Resources.Grid_Edit }) : Html.Raw("&nbsp;"), style: "action"));
```
willdean: Is `User.IsInRole(Admin)` the right test here?  What about people who have permissions on the repo but who aren't admins?

Bonobo.Git.Server/Attributes/WebAuthorizeRepositoryAttribute.cs
```diff
@@ -15,15 +15,41 @@ public class WebAuthorizeRepositoryAttribute : WebAuthorizeAttribute
...
+            //do base role checks
             base.OnAuthorization(filterContext);
 
             if (!(filterContext.Result is HttpUnauthorizedResult))
             {
-                Guid repoId;
-                var urlhelper = new UrlHelper(filterContext.RequestContext);
-                if (Guid.TryParse(filterContext.Controller.ControllerContext.RouteData.Values["id"].ToString(), out repoId))
+                if (urlhelper == null)
+                {
```
willdean: I'm not a big fan of all this `if(urlHelper == null)` stuff - is there some way to tidy this method up without all the conditions?

Bonobo.Git.Server/Attributes/WebAuthorizeRepositoryAttribute.cs
```diff
@@ -55,5 +81,22 @@ public override void OnAuthorization(AuthorizationContext filterContext)
...
             }
         }
+        private static Guid GetRepoId(AuthorizationContext filterContext)
+        {
+            Guid result;
+            if (Guid.TryParse(filterContext.Controller.ControllerContext.RouteData.Values["id"].ToString(), out result))
+            {
+                return result;
+            }
+            else
```
willdean: What situation does this condition cope with?  Do we have `Guid.Empty` to mean 'unknown repo' elsewhere?

Bonobo.Git.Server/Controllers/RepositoryController.cs
```diff
@@ -206,7 +206,7 @@ public ActionResult Delete(RepositoryDetailModel model)
             return RedirectToAction("Index");
         }
 
-        [WebAuthorizeRepository]
+        [WebAuthorizeRepository(AllowAnonymousAccessWhenRepositoryAllowsIt =true)]
         public ActionResult Detail(Guid id)
```
willdean: Spacing

Bonobo.Git.Server/Controllers/RepositoryController.cs
```diff
@@ -206,7 +206,7 @@ public ActionResult Delete(RepositoryDetailModel model)
             return RedirectToAction("Index");
         }
 
-        [WebAuthorizeRepository]
+        [WebAuthorizeRepository(AllowAnonymousAccessWhenRepositoryAllowsIt =true)]
```
willdean: Spacing around =

Bonobo.Git.Server.Test/IntegrationTests/IntegrationTestHelpers.cs
```diff
@@ -191,6 +191,10 @@ public void SetCheckbox(IWebElement field, bool select)
             if (select != field.Selected)
             {
                 field.Click();
+                if (select != field.Selected)
+                {
+                    field.SendKeys(" ");
```
willdean: What's this doing?

willdean:
Thanks for this - I have added a few comments around the place.  I am slightly concerned about the security review aspects of this - obviously we've made a bunch of stuff more permissive, and it worries me that we might have made something accessible that wasn't.

Bonobo.Git.Server.Test/IntegrationTests/IntegrationTestHelpers.cs
jeffgaroutte: There was a problem with selenium web driver clicking the remember me checkbox so I added a second check to make sure it was in the correct state and used some typing a space (" ") to make sure it was checked (or not).  

I am not certain if it was a windows10/IE11 combition issue or not, but the selenium driver was failing to check the remember me checkbox locally.

Bonobo.Git.Server/Attributes/WebAuthorizeRepositoryAttribute.cs
jeffgaroutte: There are a number of ways to do this ranging from sub-classing the the attribute to simply creating the instance at the top and always having it.  However, instantiation of a class can be expensive and if it isn't going to be used there is no need to set it - this would be the case when  AllowAnonymousAccessWhenRepositoryAllowsIt = false and  base.OnAuthorization(filterContext) sets the result to HttpUnauthorizedResult.  

We could move it into a smaller scope but if AllowAnonymousAccessWhenRepositoryAllowsIt = true and the repository does not allow anonymous access we'd be instantiating them a second time. 

I did not want to create a separate attribute when it the WebAuthorizeRepositoryAttribute seemed like the most logical place to handle a permission check of this nature.

Bonobo.Git.Server.Test/IntegrationTests/IntegrationTestHelpers.cs
willdean: OK, that's cool - maybe stick a comment in there - the Selenium-based tests are flakey, so anything that improves that is good.

Bonobo.Git.Server/Attributes/WebAuthorizeRepositoryAttribute.cs
jeffgaroutte: Previously line 26 checked if the parse was successful or not and had plenty of logic if it failed.
Line 24 declares the variable repoId without setting a value, since the default is Guid.Empty and a failed parse does not change the value this ensure the value is unchanged for what it would have been otherwise for the existing code.

Bonobo.Git.Server/Views/Repository/Index.cshtml
jeffgaroutte: Valid point.  I will switch that to use the IsCurrentUserAdministrator propety on the RepositoryDetailModel

Bonobo.Git.Server/Attributes/WebAuthorizeRepositoryAttribute.cs
kfarnung: Quick clarification, C# requires out parameters to be set before return. In this case the parse method is setting an empty GUID value rather than preserving the original value.

kfarnung:
Does this change the behavior of installs where no anonymous access is allowed? I don't want the home page to load unless the user is logged in.

Bonobo.Git.Server/Attributes/WebAuthorizeRepositoryAttribute.cs
jeffgaroutte: My mistake.

jeffgaroutte:
The repository index does load, per jakubgarfield's comments on issue #77.
Correct me if I am wrong, but anonymous pull access is configured on a per repository basis and there is not a global setting to disable it.

kfarnung:
As long as it doesn't list the private repos it should be fine. In my case there would be no private repos so it would be nice if it just redirected to the login page.

jeffgaroutte:
No, the repository index only returns repositories that the user has pull access.  In the case of anon that would only be the repositories that had anonymous access enabled.

 I thought about checking if any repository allowed Anonymous access and redirecting based on that.  However, that means the behavior of the page would be dependent on the configuration of all of the repositories.

However that logic would be in the controller method and not the attribute.

It is a bit of scope creep, but I do see value there over adding a global setting to disable anonymous access.

willdean:
I like the idea that if you have no repos with anon access, then you don't get to see the index page or anything if your anon.  It shouldn't need an extra parameter - just a test like your example above (but use Linq `.Any()`, not `Count() == 0`)  :-)

jeffgaroutte:
If I am not mistaken, this change will break the EnsureCookiePersistBetweenBrowserRestart test depending on what data is in the database for the test runner.  I have not looked into how the testing is setup enough to write a correct test or fix that test.

On a side not if I really got the code correct without an ide over lunch, I really need to get out more.

Bonobo.Git.Server/Controllers/RepositoryController.cs
```diff
@@ -39,6 +39,10 @@ public class RepositoryController : Controller
         public ActionResult Index(string sortGroup = null, string searchString = null)
         {
             var firstList = this.GetIndexModel();
+            if (!User.Identity.IsAuthenticated && !firstList.Any())
```
willdean: I know it was there already so it's not your fault, but is there a better name than 'firstList'?

Bonobo.Git.Server/Attributes/WebAuthorizeRepositoryAttribute.cs
```diff
@@ -15,15 +15,41 @@ public class WebAuthorizeRepositoryAttribute : WebAuthorizeAttribute
...
+            Guid repoId = Guid.Empty;
+            UrlHelper urlhelper = null;
+
+            // is this set to allow anon users?
+            if (AllowAnonymousAccessWhenRepositoryAllowsIt)
+            {
+                urlhelper = GetUrlHelper(filterContext);
+                repoId = GetRepoId(filterContext);
+                //if the user is authenciated or the repo id isnt there let the normal auth code handle it.
+                if (repoId!=Guid.Empty && !filterContext.HttpContext.User.Identity.IsAuthenticated)
```
kfarnung: nit: spacing around `!=`

Bonobo.Git.Server/Attributes/WebAuthorizeRepositoryAttribute.cs
```diff
@@ -15,15 +15,41 @@ public class WebAuthorizeRepositoryAttribute : WebAuthorizeAttribute
...
+                {
+                    //we are only allowing read access here.  The web ui doesnt do pushes
+                    if(RepositoryPermissionService.HasPermission(Guid.Empty,repoId, RepositoryAccessLevel.Pull))
```
kfarnung: nit: spacing after comma

jeffgaroutte:
Since this has been out for a while I figured I'd ask.  Am I missing a step to get this merged, is there a specific time that PRs get merged (like end of month), or is there a process?

willdean:
@jeffgaroutte No, there's no process, there's a vague hope that someone with the right access (probably me or @larshg ) will decide this is OK and merge it - sorry that hasn't happened sooner.

It's quite a big change over a large number of commits and I probably need to review it again properly which I will try to do.

jeffgaroutte:
Thanks for the information,

PR #721, opened 2017-07-22
Updated translations for Turkish language.

bugrakaan:
Some unneccesary commas removed, better delete confirmation text wording and proper translation of "about" page text.


bugrakaan:
I've closed this because I've just noticed one more string with unneccesary commas.

willdean:
You don't need to close a PR if you make additional changes - all your commits to the branch that the PR is from will get put into the PR - when we merge it we can decide if the get squashed together or not.

But thanks for the changes, anyway.

bugrakaan:
I had no idea and it works like this, thank you for the explanation!

PR #722, opened 2017-07-22
Updated translations for Turkish language.

bugrakaan:
Some unneccesary commas removed, better delete confirmation text wording and proper translation of "about" page text.


willdean:
Thanks!

PR #740, opened 2017-09-14
Added X-UA-Compatible meta tag

CatStarwind:
Added X-UA-Compatible meta tag with "IE-Edge" to force latest rendering mode in IE.


CatStarwind:
Simple fix for my use-case where intranet sites are set to display in compatibility mode. This meta tag forces Edge.

Sorry for making a second pull request, this is my first time doing something like this.

willdean:
Thanks! You do realise you've now volunteered to fix any compat problems that result :-)

CatStarwind:
I accept. :p

PR #763, opened 2017-11-21
Added hotfix for initial WindowsAuthentication use.

### LINKED ISSUE #743: Version 6.3.0 - Redirection After Login
Upgraded to version 6.3.0 and using Active Directory Membership Service . Having issues with the redirect after authentication similar to the issues described in #731 and #733.

On a successful login the URL in address bar changes to /Home/Logon?returnUrl=%2FBonobo.Git.Server%2FHome%2FIndex but still stuck on the login page. 

Know login was successful because I can manually put /Account in the address bar and get to the accounts index view.

CatStarwind:
Attempt to fix #743 .

### DIFF
--- a/Bonobo.Git.Server/Controllers/RepositoryController.cs
+++ b/Bonobo.Git.Server/Controllers/RepositoryController.cs
@@ -41,7 +41,7 @@ namespace Bonobo.Git.Server.Controllers
             var unorderedRepositoryDetails = GetIndexModel().ToList();
             if (!User.Identity.IsAuthenticated && !unorderedRepositoryDetails.Any())
             {
-                return RedirectToAction("Logon", "Home", new { returnUrl = Url.Action("index", "Home") });
+                return RedirectToAction((AuthenticationProvider is WindowsAuthenticationProvider ? "WindowsLogin" : "Logon"), "Home", new { returnUrl = Url.Action("index", "Home") });
             }
             if (!string.IsNullOrEmpty(searchString))
             {


willdean:
Thanks for this, but I feel like something more fundamental is wrong here - there should be a general way of bouncing people to the correct login page, which doesn't have to be reimplemented all over the place.  This is, after all, how this used to work, because the framework did the bounce using an attribute.

Could we perhaps have the `Home/Logon` action do a redirect to `WindowsLogin` if it's needed?

CatStarwind:
I believe this is the commit that introduced this bug: https://github.com/jakubgarfield/Bonobo-Git-Server/commit/a4b29666bc2f820eed058b0a39b5f51b7bf8d300#diff-542396835ce4976bb130a2c3363ed972

willdean:
@CatStarwind Yes, you're quite right, that's where it comes from - unfortunately the author of that PR doesn't appear to be around any more.

Would you be able to look at my suggestion about fixing this in the Home/Logon action rather than in RepoController/Index (which I think should have as little Auth-specific code in as possible)?

CatStarwind:
I remember trying before but being at a loss in trying to find a way without having to import an entire scope. There didn't seem to be a way to check for WindowsAuthenticationProvider in the HomeController. :\

I'll try giving it a second attempt.

CatStarwind:
Apologies for the double commit (still not accustomed to using GitHub) but would this work @willdean ?

PR #764, opened 2017-11-23
[POC] Initial dotnet core port

latop2604:
All AD stuff was disabled.
Some view are still not converted to aspnetcore.

My main setup is composed of SqlServer, Cookie Auth, internal user database.
I quickly try to use Sqlite. Be not so much.

Most custom html helpers (like WebGrid) are not available in dotnetcore. (or not found)

I will amend this commit to preserve a relative clean history.


willdean:
@latop2604 This is a huge amount of work, I am very impressed, but I don't want to give you unrealistic expectations about this getting merged into Bonobo.  

I don't think Bonobo is stable or well-tested enough (in terms of automated test) to risk a change like this, particularly as the difficult auth scenarios (AD, etc) are so difficult to test, and they're going to be the ones which are hardest in Core anyway.

You might want to discuss this with @jakubgarfield, but I think you may be better making a fork of some kind.

latop2604:
Yes of course.
I created this pull request just for information. 
It's still fun to do.

In the future, we can create a specific (experimental) branch. If someone want to contribute to. But not now.

softlion:
It seems this port is the best bet as the original author seems to have disappear. Btw AD is easy to test using Azure Ad server. And the nuget works perfectly.

Issue #299, opened 2015-07-27
HTTP Hooks


OO00O0O:
GitHub type HTTP hooks for integrations.

CoderCow:
### Open Questions

- [ ] What exactly is GitHub's "Secret" setting and is something like this useful for Bonobo hooks?
- [ ] Should hooks support http basic auth?
- [ ] Is a different format than JSON actually useful?
- [ ] How customizable should the JSON data in the request body be? AFAIK there's no standardized format for the data. Could allow to choose between "GitHub Format", "Bitbucket Format" etc. or even let users configure their own JSON templates.

# Purposed Solution (WIP)

## Triggers for Hooks

**Application Triggers**
* Repository Added
* Repository Deleted
* User Added
* User Deleted
* Team Added
* Team Removed
* Team Members Modified

Hooks for application triggers will be configured in the Bonobo settings.

**Repository Triggers**
* Branch Created
* Branch Deleted
* Branch Modified
* Tag Created
* Tag Deleted
* Repository User Access Modified (triggers for user, team or admin changes)
* Repository Anonymous Push Policity Changed

Hooks for repository triggers can be configured on a per repository basis and globally as part of the Bonobo settings. Whether the global hooks are triggered for a specific repository can be configured in each repository's settings.

## Webhooks

Multiple webhooks can be configured for each trigger. The user can 
* set a custom name and a url to invoke,
* activate and deactivate them,
* test them in the web interface and get immediate feedback,
* turn SSL certification on or off per hook.

Issue #13, opened 2011-10-08
Email Support


jakubgarfield:

jchristum:
Hello jakub. I am new to github. I found your site and implemented bonobo git server in our site for a very small team. I was wondering if the email notifications has been implemented for this? If not i'd like to take a shot at it as one of my side projects. Thank you.

jakubgarfield:
Not yet, you can give it a try :)

josephrodriguez:
With everything I read about this feature I wish I could work on it, I have some ideas about it:
- First, I think that from the site should be able to manage mail settings, to which can only be accessed by administrators, these are for example, SMTP server, port, user, password, SSL support, etc.
- The mailings should perform asynchronously, to accomplish this you could use the pattern of producer-consumer using windows queues with libraries like Rebus, NServiceBus, etc.
- Users only have to select the type of notification to which they wish to subscribe, in the configuration view.

jakubgarfield:
1) these settings could be configured in web.config - no need to bother with UI

2) true

3) that make sense, but it could be added later, firstly you can start with just one settings (emails on/off on user level)

josephrodriguez:
Ok, I'll start with these agreements and then we make the improvements.

RedX2501:
@josephrodriguez  I guess since no PRs have been posted for this i can set this up as up-for-grabs again?

PR #769, opened 2018-01-02
Reimplement the Audit Feature and Git Pack Parsing

CoderCow:
The new implementation is much slimmer and easier to grab and shouldn't cause any noticeable performance overhead at all (which the old implementation did very much).
The way I designed it should make implementing features like #13 and #299 a breeze, at least regarding the Git stuff. It can also be extended to introduce things like per-branch permissions in the future, if desired.

Compared to the current state of the codebase, the only critical thing this implementation adds is the [ReceivePackInspectStream](https://github.com/CoderCow/Bonobo-Git-Server/blob/patch-1/Bonobo.Git.Server/Git/ReceivePackInspectStream.cs) class, because an instance of it is [installed in the input stream](https://github.com/CoderCow/Bonobo-Git-Server/blob/patch-1/Bonobo.Git.Server/Git/GitService/GitHandlerInvocationService.cs#L46-L49) for every push operation. It could cause pushes to fail. There's no way but to test it throughly against different Git clients and push scenarios before we can call this production ready.
That said, nothing else added by this PR can be considered critical - the worst thing which could happen is that some commits may not have notes added to them when audit is enabled...

Note that I've removed the durability / recovery feature completely as it didn't really make sense to me. If the Git process fails, the new handler which adds the commit notes will never be invoked. If the Git process succeeds but adding notes to the commits fails for some reason, then we should try to fix the root of that problem instead of temporarily saving request data to disk and trying to add the notes over and over again until it works. 
I don't see why it was implemented like that.

I've removed these configuration settings:
* IsPushAuditEnabled
* RecoveryDataPath

Please let me know what you think about the new implementation. If it looks alright, I'll go ahead and write some tests, especially for the ReceivePackInspectStream class.


Bonobo.Git.Server/Git/GitService/GitHandlerInvocationService.cs
```diff
@@ -0,0 +1,200 @@
...
+            _next.ExecuteServiceByName(correlationId, repositoryName, serviceName, options, inspectStream, outStream);
+
+            // because the Git process has successfully finished, the request really shouldn't fail due to any exceptions thrown from here
+            try
+            {
+                Debug.WriteLine(
+                    $"{nameof(GitHandlerInvocationService)} has found {inspectStream.PackObjectCount} objects in the receive-pack stream.");
+
+                DirectoryInfo repoDirectory = _repoLocator.GetRepositoryDirectoryPath(repositoryName);
+                Debug.Assert(repoDirectory.Exists);
```
larshg: Should this throw an exception rather than being an assert? the assert is not part of a release build afaik?

larshg:
I have quickly looked through it and tested it with a few small commits plus a ~1 GB commit and it seems to work fine :)

Generally it looks very well written. So +1 for some unit test.

In addition I used git 2.15.0.windows.1 as my local client.

larshg:
Is it possible to add more handlers or is only one supporrted as is? for me it looks as only one is supported, but at places you write as  if multiple handlers could be added?

Bonobo.Git.Server/Git/GitService/GitHandlerInvocationService.cs
CoderCow: Yes, `Debug.Assert` is usually removed by the preprocessor for a release build. Because the Git process has successfully executed right before the repo directory is resvoled, I do just assume that it should exist in production.

The reason why I'm doing the assert there is to simplify debugging in case the IRepositoryLocator implementation doesn't work as expected - something which will most likely be noticed during debug / test sessions and not during production.

CoderCow:
> Generally it looks very well written.

Thank you sir.

> Is it possible to add more handlers or is only one supporrted as is?

Yep, if you wanted to add more handlers you'd just decorate one another - pretty much the same way as the `IGitService` implementations do it. This gives more control over the order and when (if even) each handler decides to call the next one.

CoderCow:
Alright, added some unit tests for the important parts of `ReceivePackInspectStream` and actually fixed two bugs. The stream is now tested against every possible vital push scenario I could think of (using an up-to-date version of Git).
Considering the already existing msys integration tests, which are using different Git client versions and perform some basic pushes, I'd call this ready for merge now.

r3dqu33n:
This could be a good starting point thinking about the usefulness of implementing webhooks.

the-Arioch:
if it had so good reviews i only con wonder why it was denied still after 2 years...

Issue #785, opened 2018-02-11
[Bug] Exceptions at startup


generik0:
There are exceptions at startup. Unneeded exceptions that can be fixed easily.
Please see pull request

PR #786, opened 2018-02-11
#785 Fix exceptions at startup

generik0:
Fix exceptions at startup

### DIFF
--- a/Bonobo.Git.Server/Configuration/ConfigurationEntry.cs
+++ b/Bonobo.Git.Server/Configuration/ConfigurationEntry.cs
@@ -27,6 +27,10 @@ namespace Bonobo.Git.Server.Configuration
                 {
                     try
                     {
+                        if(!File.Exists(_configPath))
+                        {
+                            return _current = new Entry();
+                        }
                         using (var stream = File.Open(_configPath, FileMode.Open))
                         {
                             _current = _serializer.Deserialize(stream) as Entry;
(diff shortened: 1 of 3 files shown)


generik0:
HI

I use OzCode, where you can enable clr exceptioons to break. And the changes i have made here are because excptions actually are being triggered at startup.
So all your comments are corret, but these are things actually happening. 

It is mostly because of this in global.asax at @line  183:

```c#
container.RegisterInstance(
                new GitServiceExecutorParams()
                {
                    GitPath = GetRootPath(ConfigurationManager.AppSettings["GitPath"]),
                    GitHomePath = GetRootPath(ConfigurationManager.AppSettings["GitHomePath"]),
                    RepositoriesDirPath = UserConfiguration.Current.Repositories,
                });
```

IoC is used but the static UserConfiguration.Current.Repositories is used before the initialize @lin 82

generik0:
I.e. "container.RegisterInstance(new GitServiceExecutorParam" calls before UserConfiguration.Current.Repositories is registered. 
This throws.

willdean:
Sorry, I wasn't arguing that the exceptions were happening (you can see them in VS even without OzCode, actually).

I would like us to get the best possible fix for them, though - the best fix for a null reference exception is not always `if(x != null) {  ... }` - it might be to find where the null is coming from.

Can you suggest a change to the initialisation order to prevent the `UserConfiguration.Current` problem?

generik0:
Yes it would 👍 
The only thing is, i hope within a month or so to provide a feature. To change to autofac.
Autofac is not my favourite IoC but, the code is easily portable. Making the Autofac become IoC will be the first step to a aspnetcore applicaiton.

Should we leave this merge for now, and fix the registration later?

Issue #738, opened 2017-09-14
6.2.2-6.30  Alternative UPN Suffixes Problem


devaraja666:
I have 1 Alternative UPN Suffix on the Forest, till version 6.2.1 that was not a problem. But after the update i get this error and the user are not correctly populated.
Failed to create Directory context for domain "domain2.com".
System.DirectoryServices.ActiveDirectory.ActiveDirectoryObjectNotFoundException: The specified domain does not exist or cannot be contacted.
   at System.DirectoryServices.ActiveDirectory.Domain.GetDomain(DirectoryContext context)
   at Bonobo.Git.Server.Helpers.ADHelper.GetDomain(String parsedDomainName)

For what i can see, it is trying to contact the domain suffix of the userPrincipalName, but because it is an alternative Suffiex it does not really exist as a domain inside the forest.

Im attaching the information i got from /home/diagnostics for 2.2.1 and the one after.
[log bonobo 2-2-1.txt](https://github.com/jakubgarfield/Bonobo-Git-Server/files/1303606/log.bonobo.2-2-1.txt)
[log bonobo 3-0.txt](https://github.com/jakubgarfield/Bonobo-Git-Server/files/1303607/log.bonobo.3-0.txt)

larshg:
I don't think there is an issue with the alternate suffix - it will just try to connect and if no response - try the next domain in list.

Have you removed some members or moved them from one group to another?

devaraja666:
no, i haven't removed member or edited the users. i event delete the the ADBackend folder to allow it to generate it again but it still fails as you can see on the logs it only detect 3 user on version 6.3.0 but on version 6.2.1 it does detect all the users.

i did the test with the same configuration, just upgraded it by keeping the web.config and app_data folder and deleting the rest as usual. i when back to version 6.2.2 and got the same error but on 6.2.1 it works ok.

on version 6.2.2 and 6.3.0 there where some changes to the way AD domain is handle with could explain why it is not working on later versions.

theterminator3000:
I have noticed the same issue on my installation. I deviated from the original install instructions by creating a new IIS site with SSL enabled and a custom DNS name. It would be good to get it working again as AD management is awesome!

PR #795, opened 2018-03-25
Store the actual domain username in the user model

cbaerike:
This change includes 2 things:
- in UpdateUsers(), the previous code converted UserPrincipal objects into username strings, then back into UserPrincipal objects ... losing the original UserPrincipal.Context on the way, and with that any way to determine the actual user domain.
- the UserPrincipal.UserPrincipalName property does not always contain the correct domain name after the @ character (e.g. in environments where an internal "company.local" domain was connected with Office 365, and all users received the external email address as their UPN. Using the SamAccountName and user.Context.Name instead fixes this.

### DIFF
--- a/Bonobo.Git.Server/Data/ADBackend.cs
+++ b/Bonobo.Git.Server/Data/ADBackend.cs
@@ -129,7 +129,7 @@ namespace Bonobo.Git.Server.Data
                     result = new UserModel
                     {
                         Id = user.Guid.Value,
-                        Username = user.UserPrincipalName,
+                        Username = user.SamAccountName + "@" + user.Context.Name,
                         GivenName = user.GivenName ?? String.Empty,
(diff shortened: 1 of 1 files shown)


cbaerike:
Just saw issue #738 - this PR will fix it.

gabe-alex:
Hello!

Any reason why this hasn't been merged yet?
I just installed Bonobo and am facing the same issue.

willdean:
Unfortunately this is the kind of change seems almost certain to break something in the very complex matrix of different auth permutations, many of which I'm not actually able to test here.

For example, I am somewhat sceptical that `user.SamAccountName + "@" + user.Context.Name` is a universally better way to get the name of a principal than `user.UserPrincipalName`, even though I do recognise that it probably works better for some people in some situations.

Something more adaptable is probably needed here, though I'm not really sure what.

Issue #838, opened 2019-07-03
Old year in web footer section


HerrPi:
In the footer section of the Bonobo web view it still says (c) 2015.
It should probably be 2019, instead.

PR #839, opened 2019-07-03
Fix for #838: Old year in footer section

HerrPi:
2019 is the current year now.


willdean:
Thanks!  It may be 2020 before there's another release, of course, but we'll take this anyway...

Issue #747, opened 2017-10-10
'git fetch' sporadically hangs


BlPrc:
When running 'git fetch' on repositories with a large number of refs, the git client will sometimes stall. On the server, idle git-upload-pack.exe processes remain. Once this happens, this is always reproducible the same local checkout. When fetching from a different source, the problem disappears.

ijd02:
Could you please provide more information like source file and line, config settings, error results, error logs, etc?
I'm looking into correcting source code but I'm new to Bonobo & git source.

I have an issue where a fetch operation stalls and the client reports & then hangs. Using TortoiseGit & git for windows 2.14.1.windows.1 as the client and Bonobo 5.1.1 or 6.5.0 with built-in git 1.9.5 or latest git (2.22.0.windows.1) as the server the error reported by the client is:
	git.exe fetch -v --progress "origin"

	POST git-upload-pack (gzip 1232 to 692 bytes)
	POST git-upload-pack (gzip 1432 to 793 bytes)
	POST git-upload-pack (gzip 2932 to 1542 bytes)
	POST git-upload-pack (gzip 6132 to 3148 bytes)
	POST git-upload-pack (gzip 12532 to 6347 bytes)

After this point client hangs.
On the server I see that two processes are running but with no activity, for built-in git 1.9.5 processes are git.exe and git-upload-pack.exe, when 2.22.0 is used two git.exe instances are running.
If I first fetch just master then fetch everything else no error occurs, an easy work around on my local machine but not practical on my Jenkins build infrastructure.

Ataye:
Man, i've been battling this for the last week or so!  I ended up putting some logging into the bonobo source and running my compiled dll to try to log where its happeneing etc.  I also thought to make the calls asynchronous, but I just had to get this issue resolved.  I just ran into it again tonight.  I could still clone, and after cloning again fetch is working again, but for how long?

I ended up cloning the bare repo on the server filesystem using --mirror, doing some pruning and pushed to a new repo also created with --mirror, and updating the refs so all origin become local.  Then deleted the origin remote and copying the new repo back where the old repo was.  This fixed it for most of our devs, except me tonight!

[...]

PR #840, opened 2019-07-04
Fix for #747 'git fetch' sporadically hangs, GitServiceExecutor now asynchronous.

ijd02:
Fixed git deadlock bug in GitServiceExecutor. Fault due to stdIn & stdOut buffers in the git.exe filling and blocking further processing. Handling in GitServiceExecutor now asynchronous i.e. Bonobo is continually emptying the buffers even while its sending data to git. Link to initial issue:
https://github.com/jakubgarfield/Bonobo-Git-Server/issues/747

### DIFF
--- a/Bonobo.Git.Server/Git/GitService/GitServiceExecutor.cs
+++ b/Bonobo.Git.Server/Git/GitService/GitServiceExecutor.cs
@@ -7,6 +7,8 @@ using System.Linq;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace Bonobo.Git.Server.Git.GitService
 {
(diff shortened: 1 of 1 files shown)


willdean:
Thanks for this.  There are a couple of issues I would like to discuss:  (Note that I didn't write the original code, so I'm not defensive about it in the slightest!)

* Is it necessary for both the input and the output stream to be async like this?  If we start the output read before writing the input data, then the input write could probably stay synchronous, like it used to be?

* If we're going to fix-up this broken routine, we should probably also make sure we're reading stderr at the same time as stdout, because otherwise is there still some kind of possible deadlock in the situation where a lot of stderr output occurs?  (See the last sample on https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standardoutput?view=netframework-4.8 )

* I find myself a little uncomfortable about the mixture of an old-style pre-Task API (Process), modern Task-async and then the Thread.Sleep() and the Task.Wait() call.   I wonder if it's better to avoid the CopyAsync and just use the old-style Process async methods, so avoid mixing approaches here?  This is not a fully-thought-through position, just gut feel.

But thanks for the PR, anyway, I would definitely like to add this improvement once we're fully happy with it.

ijd02:
Hi Will,I wondered about the need for input to be async, once fetch was was working I never went back and tested if synchronous would work.
Yes fixing stderr would probably wise. I focused on my immediate issue only. I'm not even sure how stderr is being handled currently.
I'm totally new to c# and do little Windows programming so the mixture of old and new styles is inadvertent and may be incorrect. While programming and looking up various API documentation I didn't come across any warning about old/new style. What exactly do you mean by "old-style Process async methods"?I did initially try using OutputDataReceived of the Process class but this wasn't viable due to newlines being mangled i.e. not suitable for raw binary data.
I didn't totally follow documentation around the async keyword and related functionality, but I'm highly experience with multi-threaded programming in assembler/C/C++ on a range of operating systems. My understanding is that calling process.StandardOutput.BaseStream.CopyToAsync() will create a new thread to execute the operation and terminate once complete. When this thread has read all available bytes from stdout the next attempt to read byte will block the thread until a new byte becomes available. i.e. the operation never terminates except when the process terminates.
The only other way I could think of was to manually create thread and write my own loop that continually calls read on stdout.  

This email is not my primary account so responses may be a little delayed, I was also away over the last few days.
Regards, Ian.

    On Friday, 5 July 2019, 08:51:10 am UTC, Will Dean <notifications@github.com> wrote:  
 
 
Thanks for this. There are a couple of issues I would like to discuss: (Note that I didn't write the original code, so I'm not defensive about it in the slightest!)
   
   -    
Is it necessary for both the input and the output stream to be async like this? If we start the output read before writing the input data, then the input write could probably stay synchronous, like it used to be?

   -    
If we're going to fix-up this broken routine, we should probably also make sure we're reading stderr at the same time as stdout, because otherwise is there still some kind of possible deadlock in the situation where a lot of stderr output occurs? (See the last sample on https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standardoutput?view=netframework-4.8 )

   -    
I find myself a little uncomfortable about the mixture of an old-style pre-Task API (Process), modern Task-async and then the Thread.Sleep() and the Task.Wait() call. I wonder if it's better to avoid the CopyAsync and just use the old-style Process async methods, so avoid mixing approaches here? This is not a fully-thought-through position, just gut feel.


But thanks for the PR, anyway, I would definitely like to add this improvement once we're fully happy with it.

Issue #835, opened 2019-06-14
[bug] Password Length max 50 Charakters


m4m-simon:
I've changed my admin Passwort to a new 64 Charakter Password.
When i tried to log back in with the new Passwort, it says that only 50 Charakters are allowed. So I lost my access to the system!!!
Luckily i was able to restore a backup of the "Bonobo.Git.Server.db" from few hours before.

digitalcoyote:
I  had a couple of users run into this issue as well, If I get some time, I'll look into getting a PR together to fix it.

digitalcoyote:
I think I may have correctly added the validation, but I needed to boot into Windows (working with Rider on Linux Mint) and I've got so many updates that have to install before it will be usable... I'll try to test my code tomorro

PR #850, opened 2019-09-29
Fix for  #835

digitalcoyote:
Added StringLength validator to prevent saving passwords longer than 50


willdean:
This is a fix for #835

willdean:
Thanks!

PR #873, opened 2020-07-22
Add AD Caching - Fixes #872

### LINKED ISSUE #872: AD Integration - Performance is quite slow due to lots of repeated queries to AD
This is purely a performance improvement for when AD authentication is invoked.

Currently, whenever anything that involves AD is triggered, lots of repeated calls to `UserPrincipal.FindByIdentity`, `GroupPrincipal.FindByIdentity`, and `GroupPrincipal.GetMembers(true)` are made. In addition to that, `PrincipalContext` objects are spawned and killed off a bunch, even during a single operation (effectively opening a context with AD once per query).

This all causes AD code to run pretty slowly: in our environment literally any Git method that talks to the remote takes upwards of 15 seconds to complete (`git pull`, `git push`, `git remote prune` etc.).

rozniak:
This PR introduces support for caching AD contexts, principals, and group member queries.

Two config keys are added:
`ActiveDirectoryGroupQueryCacheExpiry` - the cached results of group member enumerations will be considered stale after this amount of time
`ActiveDirectoryPrincipalCacheExpiry` - the cached `GroupPrincipal` and `UserPrincipal` objects will be considered stale after this amount of time

Fixes #872

### DIFF
--- a/Bonobo.Git.Server/Data/ADCachedPrincipal.cs
+++ b/Bonobo.Git.Server/Data/ADCachedPrincipal.cs
@@ -13,13 +13,13 @@ namespace Bonobo.Git.Server.Data
         public PrincipalContext PrincipalContext { get; private set; }
 
 
-        private bool Disposing;
+        private bool _Disposing;
 
 
         public ADCachedPrincipal(PrincipalContext pc, Principal principal)
         {
+            _Disposing = false;
             CacheTime = DateTime.UtcNow;
-            Disposing = false;
             Principal = principal;
             PrincipalContext = pc;
         }
@@ -27,12 +27,12 @@ namespace Bonobo.Git.Server.Data
 
         public void Dispose()
         {
-            if (Disposing)
+            if (_Disposing)
             {
-                throw new ObjectDisposedException("Principal");
+                throw new ObjectDisposedException(nameof(Principal));
             }
 
-            Disposing = true;
+            _Disposing = true;
             Principal.Dispose();
         }
     }
--- a/Bonobo.Git.Server/Data/ADCachedPrincipal.cs
+++ b/Bonobo.Git.Server/Data/ADCachedPrincipal.cs
@@ -1,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.DirectoryServices.AccountManagement;
+using System.Linq;
+using System.Web;
+
+namespace Bonobo.Git.Server.Data
+{
+    public class ADCachedPrincipal
+    {
+        public DateTime CacheTime { get; private set; }
+        public Principal Principal { get; private set; }
+        public PrincipalContext PrincipalContext { get; private set; }
+
+
+        public ADCachedPrincipal(PrincipalContext pc, Principal principal)
+        {
+            CacheTime = DateTime.UtcNow;
+            Principal = principal;
+            PrincipalContext = pc;
+        }
+    }
+}
--- a/Bonobo.Git.Server/Data/ADCachedPrincipal.cs
+++ b/Bonobo.Git.Server/Data/ADCachedPrincipal.cs
@@ -13,12 +13,12 @@ namespace Bonobo.Git.Server.Data
         public PrincipalContext PrincipalContext { get; private set; }
 
 
-        private bool _Disposing;
+        private bool _disposing;
 
 
         public ADCachedPrincipal(PrincipalContext pc, Principal principal)
         {
-            _Disposing = false;
+            _disposing = false;
             CacheTime = DateTime.UtcNow;
             Principal = principal;
             PrincipalContext = pc;
@@ -27,12 +27,12 @@ namespace Bonobo.Git.Server.Data
 
         public void Dispose()
         {
-            if (_Disposing)
+            if (_disposing)
             {
-                throw new ObjectDisposedException(nameof(Principal));
+                throw new ObjectDisposedException(nameof(ADCachedPrincipal));
             }
 
-            _Disposing = true;
+            _disposing = true;
             Principal.Dispose();
         }
     }
--- a/Bonobo.Git.Server/Data/ADCachedPrincipal.cs
+++ b/Bonobo.Git.Server/Data/ADCachedPrincipal.cs
@@ -6,18 +6,34 @@ using System.Web;
 
 namespace Bonobo.Git.Server.Data
 {
-    public class ADCachedPrincipal
+    public class ADCachedPrincipal : IDisposable
     {
         public DateTime CacheTime { get; private set; }
         public Principal Principal { get; private set; }
         public PrincipalContext PrincipalContext { get; private set; }
 
 
+        private bool Disposing;
+
+
         public ADCachedPrincipal(PrincipalContext pc, Principal principal)
         {
             CacheTime = DateTime.UtcNow;
+            Disposing = false;
             Principal = principal;
             PrincipalContext = pc;
+        }
+
+
+        public void Dispose()
+        {
+            if (Disposing)
+            {
+                throw new ObjectDisposedException("Principal");
+            }
+
+            Disposing = true;
+            Principal.Dispose();
         }
     }
 }
--- a/Bonobo.Git.Server/Data/ADCachedSearchResult.cs
+++ b/Bonobo.Git.Server/Data/ADCachedSearchResult.cs
@@ -13,13 +13,13 @@ namespace Bonobo.Git.Server.Data
         public PrincipalSearchResult<Principal> SearchResults { get; private set; }
 
 
-        private bool Disposing;
+        private bool _Disposing;
 
 
         public ADCachedSearchResult(PrincipalSearchResult<Principal> searchResults)
         {
+            _Disposing = false;
             CacheTime = DateTime.UtcNow;
-            Disposing = false;
             SearchResults = searchResults;
 
             // Search results are lazy loaded, we should enum them now so the cache is
@@ -38,12 +38,12 @@ namespace Bonobo.Git.Server.Data
 
         public void Dispose()
         {
-            if (Disposing)
+            if (_Disposing)
             {
-                throw new ObjectDisposedException("SearchResults");
+                throw new ObjectDisposedException(nameof(SearchResults));
             }
 
-            Disposing = true;
+            _Disposing = true;
 
             foreach (Principal principal in Principals)
             {
--- a/Bonobo.Git.Server/Data/ADCachedSearchResult.cs
+++ b/Bonobo.Git.Server/Data/ADCachedSearchResult.cs
@@ -6,16 +6,20 @@ using System.Web;
 
 namespace Bonobo.Git.Server.Data
 {
-    public class ADCachedSearchResult
+    public class ADCachedSearchResult : IDisposable
     {
         public DateTime CacheTime { get; private set; }
         public IList<Principal> Principals { get; private set; }
         public PrincipalSearchResult<Principal> SearchResults { get; private set; }
 
 
+        private bool Disposing;
+
+
         public ADCachedSearchResult(PrincipalSearchResult<Principal> searchResults)
         {
             CacheTime = DateTime.UtcNow;
+            Disposing = false;
             SearchResults = searchResults;
 
             // Search results are lazy loaded, we should enum them now so the cache is
@@ -29,6 +33,24 @@ namespace Bonobo.Git.Server.Data
             }
 
             Principals = principals.AsReadOnly();
+        }
+
+
+        public void Dispose()
+        {
+            if (Disposing)
+            {
+                throw new ObjectDisposedException("SearchResults");
+            }
+
+            Disposing = true;
+
+            foreach (Principal principal in Principals)
+            {
+                principal.Dispose();
+            }
+
+            SearchResults.Dispose();
         }
     }
 }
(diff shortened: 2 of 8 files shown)


Bonobo.Git.Server/Helpers/ADHelper.cs
```diff
@@ -274,5 +316,104 @@ public static PrincipalContext GetPrincipalGroup(string name, out GroupPrincipal
...
+
+            Log.Verbose("ADCACHE: Cache miss for {Guid}", guid);
+
+            return null;
+        }
+
+        private static ADCachedPrincipal GetCachedPrincipalData(string name)
+        {
+            TimeSpan tenMinutes = new TimeSpan(0, 10, 0);
+            bool     searchUpn  = name.Contains("@");
```
willdean: TimeSpan.FromMinutes ?

Bonobo.Git.Server/Helpers/ADHelper.cs
```diff
@@ -1,15 +1,24 @@
...
 namespace Bonobo.Git.Server.Helpers
 {
     public static class ADHelper
     {
+        private static Dictionary<string, PrincipalContext> CachedPrincipalContexts = new Dictionary<string, PrincipalContext>();
+        private static object CachedPrincipalContextsLock = new object();
+
+        private static List<ADCachedPrincipal> CachedPrincipals = new List<ADCachedPrincipal>();
+        private static object CachedPrincipalsLock = new object();
+
```
willdean, line 23: Rather than have to worry about locks and the possibility of deadlocks, what about using ConcurrentDictionary and something else ConcurrentXXX for CachedPrincipals

Bonobo.Git.Server/Helpers/ADHelper.cs
```diff
@@ -163,6 +171,11 @@ public static UserPrincipal GetUserPrincipal(string username)
 
             Log.Verbose("GetUserPrincipal: username {UserName}, domain {DomainName}, stripped {StrippedUserName}", username, parsedDomainName, strippedUsername);
 
+            var cachedUserData = GetCachedPrincipalData(username);
+
+            if (cachedUserData != null)
+                return cachedUserData.Principal as UserPrincipal;
```
willdean, line 183: Braces

Bonobo.Git.Server/Helpers/ADHelper.cs
```diff
@@ -212,17 +236,24 @@ private static UserPrincipal GetUserPrincipal(Domain domain, string fullUsername
         /// <returns>The Userprincipal if found, else null</returns>
         public static UserPrincipal GetUserPrincipal(Guid id)
         {
+            var cachedUser = GetCachedPrincipalData(id);
+
+            if (cachedUser != null)
+                return cachedUser.Principal as UserPrincipal;
```
willdean, line 248: Braces

Bonobo.Git.Server/Helpers/ADHelper.cs
rozniak: The last commit should replace the hard-coded `tenMinutes` with a call to parse the `TimeSpan` from the `web.config`...

Bonobo.Git.Server/Helpers/ADHelper.cs
```diff
@@ -274,5 +316,104 @@ public static PrincipalContext GetPrincipalGroup(string name, out GroupPrincipal
             }
             throw new ArgumentException("Could not find principal group: " + name);
         }
+
+
+        private static PrincipalContext GetPrincipalContext(ContextType contextType, string name)
+        {
+            string safeName = name.ToLower();
```
willdean, line 369: Is ToLower the right version of this?  Are there any issues about invariant cultures, ordinals, Turkish "i" chars, etc?

Bonobo.Git.Server/Helpers/ADHelper.cs
```diff
@@ -255,6 +286,14 @@ public static PrincipalContext GetMembersGroup(out GroupPrincipal group)
         /// <returns>Principal context on which the group was found.</returns>
         public static PrincipalContext GetPrincipalGroup(string name, out GroupPrincipal group)
         {
+            var cachedGroupData = GetCachedPrincipalData(name);
```
willdean, line 295: Would it be possible to be more precise than "name"?   I guess this could be groupName, and there may be places that could be domainName elsewhere?

Bonobo.Git.Server/Security/ADMembershipService.cs
```diff
@@ -87,10 +84,10 @@ public IList<UserModel> GetAllUsers()
...
         public UserModel GetUserModel(string username)
         {
-            using (var upc = ADHelper.GetUserPrincipal(username))
-            {
-                return ADBackend.Instance.Users.FirstOrDefault(n => n.Id == upc.Guid.Value);
-            }
+            UserPrincipal upc = ADHelper.GetUserPrincipal(username);
+
+            return ADBackend.Instance.Users.FirstOrDefault(n => n.Id == upc.Guid.Value);
+
```
willdean, line 90: Is doesn't look like this throw is reachable (this is not a bug you've introduced) - am I right?

Bonobo.Git.Server/Helpers/ADHelper.cs
rozniak: `ConcurrentBag<T>` perhaps?

Bonobo.Git.Server/Helpers/ADHelper.cs
rozniak: The input should be a domain name, don't they use Punycode for encoding non-ASCII characters? I don't *think* it should be a problem if so...

Bonobo.Git.Server/Helpers/ADHelper.cs
willdean: @rozniak - this is a good example of the multiple commits to one PR problem I was hoping to avoid - if you have one commit that replaces code in an earlier commit, it makes it really hard to review, because the GH review process works one commit at a time, and so I end-up reviewing code which you have already replaced.

It's OK having multiple commits, but not when they're changing the same code.

Bonobo.Git.Server/Helpers/ADHelper.cs
willdean: Yes, if you we don't care about ordering, which I wasn't sure about at the time I wrote that.

Bonobo.Git.Server/Helpers/ADHelper.cs
rozniak: Could do, I used the existing parameter `name`.

willdean: Oh sorry, yes - if you had improved it I could have moaned about "unrelated changes" instead.  No worries.

Bonobo.Git.Server/Security/ADMembershipService.cs
rozniak: That's what I thought when I saw it, VS wasn't complaining about it being unreachable so I decided to leave it just in case. I don't see why it would be reachable though, tempted to bin it.

Bonobo.Git.Server/Helpers/ADHelper.cs
rozniak: I see - I'll bear that mind in future.

willdean:
Thanks very much for this.  If you could look at changing to the Concurrent collections, which avoid the need for locks, that would be great.

rozniak:
No worries, I'll get all the changes sorted now. :+1:

Bonobo.Git.Server/Helpers/ADHelper.cs
rozniak: Actually `BlockingCollection<T>` might be more appropriate (I think) since we don't want to remove items unless expiring them from the cache.

willdean:
When you've got it how you want it, you can squash it all at your end into one commit and then force-push the same branch (feels like something you're not supposed to do, but it's OK), and that will supersede everything in here without you needing to make a new PR or anything.

Bonobo.Git.Server/Helpers/ADHelper.cs
willdean: I think you were probably right with ConcurrentBag - BlockingCollection is a producer/consumer thing.

rozniak: Well, I have mostly been looking at replacing the locks with `ConcurrentDictionary<TKey, TValue>` objects - the thing is I'm thinking about what happens during the logic between adding/removing values.

This kind of thing is in my mind right now as a plausible issue:
```
ADCachedPrincipal cachedResult;

if (CachedPrincipals.TryGetValue(guid, out cachedResult))
{
    if (cacheExpired)
    {
        //
        // Another cache hit could occur here and already expire the object
        // and add a new one, surely?
        //

        if (CachedPrincipals.TryRemove(guid, out cachedResult))
        {
            // disposal
        }
    }
}

// get a new one... TryAdd() etc.
```

With the locks the whole thing would be wrapped, instead of just the collection accesses?

willdean: Oh yes, I see what you mean - tricky.  But it does seem like holding a (system-wide) lock for the whole time that we're talking to AD is also a bit grim.  Not that I really think Bonobo needs to to pretend to be some kind of high-throughput web server, so maybe that doesn't matter, though clearly deadlocks would be problem.

I somehow feel that if one is prepared to tolerate the occasional situation where two callers both go off to AD to refresh the cache, but one of them loses loses the race and immediately disposes the results it's just fetched, then there may be some combination of TryGet, AddOrUpdate, TryRemove which might work.  But I can't currently work out what that is...

willdean: Is this relevant?  http://reedcopsey.com/2011/01/16/concurrentdictionarytkeytvalue-used-with-lazyt/

rozniak: I dunno, the lock would be alright would it not? Since other threads would only wait whilst a new object is being retrieved by the winner, then when that's done they'd all use the cached version... surely without locks the race condition would mean more waiting because multiple threads could make the same call to AD at slightly different times?

willdean: Yes, I think you're probably right that it's the least of the evils.

rozniak: I will tidy up the locks a bit anyway, I have noticed that I add the new object into the cache outside the lock!

Also will look into how to handle caching when it comes to `UpdateUsers()`... really that function should override the cache when enumerating group members.

Bonobo.Git.Server/Data/ADCachedPrincipal.cs
```diff
@@ -6,18 +6,34 @@
...
 {
-    public class ADCachedPrincipal
+    public class ADCachedPrincipal : IDisposable
     {
         public DateTime CacheTime { get; private set; }
         public Principal Principal { get; private set; }
         public PrincipalContext PrincipalContext { get; private set; }
 
 
+        private bool Disposing;
```
willdean, line 16: I'm sorry that the project is not 100% consistent on this, but can we use `private bool _disposing` -style names for private members, please?

Bonobo.Git.Server/Data/ADCachedSearchResult.cs
```diff
@@ -6,16 +6,20 @@
...
 {
-    public class ADCachedSearchResult
+    public class ADCachedSearchResult : IDisposable
     {
         public DateTime CacheTime { get; private set; }
         public IList<Principal> Principals { get; private set; }
         public PrincipalSearchResult<Principal> SearchResults { get; private set; }
 
 
+        private bool Disposing;
```
willdean, line 16: As above, _disposing

Bonobo.Git.Server/Data/ADCachedPrincipal.cs
```diff
@@ -6,18 +6,34 @@
...
             Principal = principal;
             PrincipalContext = pc;
         }
+
+
+        public void Dispose()
+        {
+            if (Disposing)
+            {
+                throw new ObjectDisposedException("Principal");
```
willdean, line 32: Might help a little with exception traces to have the actual class name passed into these?  Use nameof() perhaps?

Bonobo.Git.Server/Data/ADCachedSearchResult.cs
```diff
@@ -30,5 +34,23 @@ public ADCachedSearchResult(PrincipalSearchResult<Principal> searchResults)
 
             Principals = principals.AsReadOnly();
         }
+
+
+        public void Dispose()
+        {
+            if (Disposing)
+            {
+                throw new ObjectDisposedException("SearchResults");
```
willdean, line 43: As above

willdean:
Thanks.  Do you think this is safe to merge?

rozniak:
If it helps, I'm running this build internally right now without any hiccups.

Also thinking about the `throw new ObjectDisposedException`... should that refer to the property in the class that `Dispose()` disposes of, or the class itself?

rozniak:
Oh, and I know I change the private member name to `_Disposed`, should it be `camelCase` like `_disposed` ?

willdean:
It should be like my review comment ;-)  "_disposed".   I was actually just going to quietly change it after I'd merged it because I don't want to exhaust your enthusiasm with too much nitpicking.

I ought to be able to just point to the .net corefx coding standard or something and just ask people to use that.

And yes, I think the ObjectDisposedException should probably have the name of the object which is doing the complaining, not anything else within it.

PR #880, opened 2021-01-26
Add Controller Tests, fix #659 (1ef6921) and other features listed below

eugeniomiro:
This PR adds the following features

* 110 Controller tests added (2 skipped).
  * These test methods are testing the current behavior of Controllers in particular, so they we can start refactoring with a Safety Net
  * There are two Ignored tests in the list, that mark two refactoring points to create new infrastructure code to avoid direct DbContext dependencies
  * Not all controllers are fully tested yet, but I'm in the process of creating all required tests for that
* Solves #659
  I installed a server and used it with my team for our internal development projects and found that, when using Remote button in Sourcetree, the repository was not found. This PR includes a fix for that, together with tests covering the new code
* Updates Unity package to 5.11.7
* Moves all Unit tests to the Unit namespace
* Refactors ConfigurationEntry for testability
* Adds .editorconfig file
* Unity configuration code was moved into App_Start/UnityConfig class
* Test framework was migrated to MSTestV2

PR #881, opened 2021-01-26
Infrastructure Update to improve Testability

eugeniomiro:
This PR adds the following features

* Unity upgrade to 5.11.7
* Coalesce all Unit Tests into one namespace
* Test Framework Migration to MSTestV2
* refactor ConfigurationEntry for Testability


eugeniomiro:
Hi @willdean, is the project still accepting contributions? 
Is there anything I can do to improve this PR to make it acceptable?
I know you mentioned in my previous PR(#880) that there were too many changes on it but this is a smaller one, do you think I need to shrink it even more?

willdean:
Hi @eugeniomiro - sorry for not responding to this - I have not been thinking much about Bonobo recently.
I can take this - it only affects the tests.  

As I side note, and it's not going to be a problem now, but I would say that #3d283cd shows a problem where VS and nuget disagree about XML formatting and hundreds of lines get meaningless white-space differences.

In theory, every single changed line has to be looked at by someone taking a contribution, so avoiding big whitespace changes is helpful.  Obviously this is not your fault, it's VS stupidity, but just something to think about.

eugeniomiro:
Thank you @willdean, I took note of your comment about the XML in #3d283dc, will be more careful next time

willdean:
Many thanks for the contributions, I didn't say that last time.

eugeniomiro:
One more thing @willdean, I'm preparing the next PR with all those tests I sent in #880, trying not to touch the main code, that's in preparation for the fix of #882 which is the Issue I found and wanted to fix in the first place, but wanted to do it using a set of tests as a contention net. I have those in a large set of commits, is it best for you that I squash them all for the PR?
Any other suggestion for it?

Another thing, I'm adding ``` .editorconfig``` as one of my files, to keep source code formatting consistency. I leave this comment to check if you want that in or not, I think it could be very useful.

willdean:
Hi @eugeniomiro I think it probably is easier if they're squashed, so long as they're all related to each other.   One of the problems if they're _not_ squashed is that the same files often appear in multiple commits, which makes a lot more work to review.

Personally I'm ok with adding .editorconfig as long as it keeps the style consistent with the existing code.   That should probably be in a separate commit/PR though, because it's not really related to changing tests or fixing bugs.

